PR #16, opened 2017-03-02
Add maven installation in guide

MarcinMoskala:


holgerbrandl:
Thanks for pointing it out. I somehow took it for granted, but I totally agree that we should not.

To prevent that users get confused by potentially cryptic error messages, I've also created #17.

PR #38, opened 2017-08-11
$KOTLIN_OPTS overrides //KOTLIN_OPTS in scriptlet.

stephenbyrne-mfj:
I wanted to be able to change Kotlin opts (usually to add -J-agentlib:jdwp...) without changing the script, so I made $KOTLIN_OPTS in the environment override //KOTLIN_OPTS in the scriptlet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,7 @@ The intial shebang could be left out by running the script directly with `kscrip
 println("Hello from Kotlin with 5g of heap memory running in server mode!")
 ```
 
+If environmental variable `KOTLIN_OPTS` is set, it will override the value set in the scriptlet.
 
 ### Scripts with a main method do not run with `kscript`?
 
--- a/kscript
+++ b/kscript
@@ -233,7 +233,11 @@ if [ -n "$dependencies" ] && [ -z "$classpath" ];then
 fi
 
 ## Extract kotlin arguments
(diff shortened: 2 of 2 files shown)


holgerbrandl:
Thanks for the PR.

I totally see and agree to the point that a user may want to override the //KOTLIN_OPTS directive to modulate the jvm invocation.

However, I'm not sure about the proposed technical realization.
Until KOTLIN_OPTS has been in an internal directive in kscriptlets. By using it outside of `kscript` is should potentially use a more descriptive name. Something hinting its relation to `kscript`. The current name implies that it's generally about kotlin, which it is not.

To me it would feel technically more sound to follow the approach described in https://stackoverflow.com/questions/38197406/overriding-configuration-with-environment-variables-in-typesafe-config
So inside a kscriptlet we would do
```
export MY_EXPORTED_ENV_OPTS=-Xmx5g

kscript '
//KOTLIN_OPTS ${MY_EXPORTED_ENV_OPTS}
println("foo")
...
```
And since it's bash you could even provide defaults `${MY_EXPORTED_ENV_OPTS:--Xmx5g}`

This does not work yet with the current master I think, but we could make it work easily, and it should solve your problem, wouldn't it?

Imho this would be conceptually more clean, as it is making the external configuration aspect more explicit.

PR #46, opened 2017-10-31
Set jarPath correctly if $0 is a symlink.

stephenbyrne-mfj:
If kscript on the $PATH is a symlink, jarPath is not set correctly, and you get "error: could not find or load main class kscript.app.KscriptKt". This change fixes that.


holgerbrandl:
Thanks for the PR stephen. Unfortunately readlink is behaving differently on macos. See http://stackoverflow.com/questions/3915040/bash-fish-command-to-print-absolute-path-to-a-file

What we could do to fix it might be something like
```
abspath() {
    if [[ $OSTYPE != darwin* ]]; then minusFarg="-f"; fi
    echo "$(readlink ${minusFarg} $1)";
}
jarPath=$(dirname $(abspath -f $0))/kscript.jar
```
This should work for linux, macos, and cygwin. But maybe you know a better way?

stephenbyrne-mfj:
I will investigate, and make sure it works on cygwin too.

PR #47, opened 2017-10-31
Append $KSCRIPT_KOTLIN_OPTS to kotlin runtime opts.

stephenbyrne-mfj:
A way to override //KOTLIN_OPTS lines in a script.

Supersedes #38.


holgerbrandl:
Thanks. Maybe you have an idea for a simple unit-test as well, to make sure that I don't kill the feature by accident?

For existing tests see integration tests https://github.com/holgerbrandl/kscript/blob/master/test/test_suite.sh or jvm-level unit tests https://github.com/holgerbrandl/kscript/blob/master/src/test/kotlin/Tests.kt

stephenbyrne-mfj:
Commit 9f09f0b breaks it - it loses the space before the kscriptOpts.

PR #48, opened 2017-11-02
Fix space handling for adding KSCRIPT_KOTLIN_OPTS

stephenbyrne-mfj:
This is a more idiomatic way to handle the spaces. Fixes bug introduced in https://github.com/holgerbrandl/kscript/commit/9f09f0b7c8b590f451446afb356605ee73f466bc (see #47).


holgerbrandl:
Indeed, stupid me! Though adding an extension method to instead of plus().trim seems a bit bloated to me.

BTW are you at kotlinconf so that we could have an actual chat?

PR #68, opened 2017-11-22
Handle symlinks to kscript and $KSCRIPT_HOME.

stephenbyrne-mfj:


holgerbrandl:
I think the PR is obsolete because an alternative (imho more readable) method to resolve symlinks has been implemented in `master`.

PR #79, opened 2017-12-07
Add instructions for idea command line launcher

elihart:
The Intellij support is great, but I had to do some external googling after initially getting the error
```
/Users/eli_hart/.sdkman/candidates/kscript/current/bin/kscript: line 13: exec: idea: not found
```

I thought it would be nice to make this explicit.


holgerbrandl:
Good idea, thanks for PR.

PR #86, opened 2018-01-08
Simplify Gradle config for script boostraping with IDEA

gildor:


holgerbrandl:
Thanks!

gildor:
Oh, forgot to send my comment about rationale of changes, just for your information:

1. Kotlin plugin now uses plugins dsl instead of legacy builscript dsl, it's recommended way to manage plugins and requires much less code in config. Also you don't need to apply plugin manually. Plugins dsl allows to specify plugin dependecy and apply it at the same time
2. Version of stdlib now implicit (feature of Kotlin Gradle Plugin) to rid of variable with version of Kotlin (now you specify it only to plugin)
3. Removed unnecessary id and group. You need those properties only if you publish artefact, but apparently you don't have a proper way to publish script and hardcoded group and version is not better than unspecified
4. Just compact version of sourceSet configuration

holgerbrandl:
Too late! :-) 

I figured most of these points from browsing the changes, but thanks for the clarification.

It looks so much more concise now!

Issue #103, opened 2018-02-15
--idea not usable with //INCLUDE


EugeneSusla:
I really like the idea of having a dedicated `utils.kts` with common functions used by other scripts, and it's currently doable with the include directive, however if I then try to generate the `--idea` project, the file being opened is some temporary consolidated script, as opposed to a symlink to my original file as I get if I don't use includes.

I believe this can be made more useful by instead symlinking both the script and all it's included scripts recursively, while consolidating their dependencies in build.gradle. This should resolve the "red references" problem, as well as let modify both script and its dependencies all in one intellij project, making it smoother to move stuff back and forth.

I'm happy to implement and send PR if there's no objections to this approach

holgerbrandl:
Indeed, temporary projects will be based on the pre-processed script. 

That's a tricky one fix. Either we resolve them or good code will be red in the temporary intellij project. Resolving them temporary will work for some but won't for others. Eg consider:

```
//INCLUDE ../../../foo.kt
//INCLUDE https://raw.githubusercontent.com/holgerbrandl/kscript/master/test/resources/includes/include_3.kt
```
For those ones I don't see any sensible way to create symlinks. I think we need to actually resolve them to provide a temporary project context. 
Because how would you symlink them without rewriting the user script?

I guess we could fix some cases, but an incomplete solution might be more confusing than helpful. What do you think?

EugeneSusla:
For links, it should be ok to symlink to it's cached .kts file in ~/.kscript dir, like the one that gets created when calling kscript with that url directly.

For parent dirs in path - as long as we don't attempt to have the directory structure in src/ mimic that of the corresponding scripts we should be fine. The script and foo.kt would just be symlinked side by side in src/

So, for your example, the way I imagine it, the ~/.kscript would look something like this:
```
~/.kscript
├── kscript_tmp_project__url_cache_d8f22e323c8720a2.kts_1519196345122/
│   ├── build.gradle
│   ├── local.properties
│   └── src/
│       ├── actualScript.kts -> /tmp/1/2/3/actualScript.kts
│       ├── foo.kts -> /tmp/foo.kt
│       └── url_cache_d8f22e323c8720a2.kts -> ~/.kscript/url_cache_d8f22e323c8720a2.kts
└── url_cache_d8f22e323c8720a2.kts
```
[...]

PR #105, opened 2018-02-28
Support --idea with includes (#103)

EugeneSusla:
Caveat: only items declared in .kt files (not .kts) are picked up
by idea when resolving terms in its neighbours

Test:
- create a file that `//INCLUDE`s a relative .kt script, and a .kt url
- pass it to `kscript --idea` and ensure all terms from its includes are non-red
- execute it and ensure no compilation/runtime errors

### DIFF
--- a/src/main/kotlin/kscript/app/AppHelpers.kt
+++ b/src/main/kotlin/kscript/app/AppHelpers.kt
@@ -99,6 +99,9 @@ object ShellUtils {
 fun infoMsg(msg: String) = System.err.println("[kscript] " + msg)
 
 
+fun warnMsg(msg: String) = System.err.println("[kscript] [WARN] " + msg)
+
+
 fun errorMsg(msg: String) = System.err.println("[kscript] [ERROR] " + msg)
 
 
@@ -133,9 +136,10 @@ fun createTmpScript(scriptText: String, extension: String = "kts"): File {
 fun String.stripShebang(): String = lines().dropWhile { it.startsWith("#!/") }.joinToString("\n")
 
 
-fun fetchFromURL(scriptURL: String): File? {
+fun fetchFromURL(scriptURL: String): File {
     val urlHash = md5(scriptURL)
-    val urlCache = File(KSCRIPT_CACHE_DIR, "/url_cache_${urlHash}.kts")
+    val urlExtension = if (scriptURL.endsWith(".kt")) "kt" else "kts"
+    val urlCache = File(KSCRIPT_CACHE_DIR, "/url_cache_${urlHash}.$urlExtension")
 
     if (!urlCache.isFile) {
         urlCache.writeText(URL(scriptURL).readText())
@@ -190,10 +194,10 @@ fun numLines(str: String) = str.split("\r\n|\r|\n".toRegex()).dropLastWhile { it
 fun info(msg: String) = System.err.println(msg)
 
 
-fun launchIdeaWithKscriptlet(scriptFile: File, dependencies: List<String>, customRepos: List<MavenRepo>): String {
+fun launchIdeaWithKscriptlet(script: Script): String {
     requireInPath("idea", "Could not find 'idea' in your PATH. It can be created in IntelliJ under `Tools -> Create Command-line Launcher`")
 
-    System.err.println("Setting up idea project from ${scriptFile}")
+    System.err.println("Setting up idea project from ${script.file}")
 
     //    val tmpProjectDir = createTempDir("edit_kscript", suffix="")
     //            .run { File(this, "kscript_tmp_project") }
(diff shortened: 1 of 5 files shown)


holgerbrandl:
Imho the PR mixes `//include` management too much into the `Script` helper class, and the patch seemed a bit too complicated in my taste.

I've pushed a more simplistic fix into the branch https://github.com/holgerbrandl/kscript/tree/alternative_impl_pr105

What do you think about it?

The main functional difference is, that we actually need to resolve text mode and custom preamable in ant case, and if so use a temp-script in the `--idea` mode. Otherwise good code would be red, because preambles are not just methods but may also declare variables. And so does the text-processing mode. E.g. try/see the bundled test dsl with
```
/test/resources/custom_dsl/mydsl --idea test/resources/custom_dsl/mydsl_test_with_deps.kts
```

> Caveat: only items declared in .kt files (not .kts) are picked up
That's by design of Kotlin (and not related to kscript). `kts` scripts can not be used to include method definitions.

EugeneSusla:
The simplified fix looks good, apart from the preambles treatment you bring up.

With regards to that - I'm not sure I understand your point why do we need to open temp script for the case with preambles/custom DSLs/text mode. I tried running:
`/test/resources/custom_dsl/mydsl --idea test/resources/custom_dsl/mydsl_test_with_deps.kts` against this patch, and it opens in IDE with zero red, including the `foo` reference (which I'm assuming is what you're referring to when you say there can also be variables, not just functions). With additional `-t` and `lines` usage, `mydsl_test_with_deps.kts` is still non-red. What ends up red is basically anything in the preambles that assumes to be part of a `.kts`, such as top-level code and `args` references, with the former being easy to circumvent with something like
```kotlin
val ignored = run {
    //your formerly-top-level code
}
```
This is still a lot better than just not being able to edit such scripts in IDE at all.

You can achieve the same in your patch as well if you pass `scriptResource` instead of `rawScript` to `launchIdeaWithKscriptlet`, and conditionally perpend `Script(correspondingPreambleStr.lines(), "kt").createTmpScript().toURL()` to its `includeURLs` parameter for custom preamble/text mode cases.

holgerbrandl:
>>  I tried running against this patch
I think you're not running it correctly. `mydsl` is calling `kscript` internally, and I could imagine that it is not the devel version which is in your PATH? 

When using alternative_impl_pr105 it looks for me like

So no red code and the preamble is correctly inlined.

I've added another test in 9277610 to prevent regressions of this feature.

> Edit: I guess not pass scriptResource directly instead of rawScript, as that wouldn't handle urls/stdin, but rather move the preamble-appending part from prepareScript to postProcessScript, and use the latter in place of resolveIncludes

From what I can see there should not be any situation in which  a script that runs with `kscript` has red code in temp projects created using `--idea`. Could you provide a counter-example?

holgerbrandl:
We could potentially wrap the resolved dsl preamble with some comment marker to make it more clear to the user how the temp script emerged. Mockup

```kotlin
// DSL_PREAMBLE START
fun include_1() = println("include_1")
// DSL_PREAMBLE END

include_1()
```

Furthermore we could even think about adding a small README.md to temp-projects detailing out the symlinking scheme. At the moment even if all code is green, users might be a bit lost. It could also not that for temp scripts, changes need to be synced back manually into the original source.

EugeneSusla:
> I think you're not running it correctly. mydsl is calling kscript internally, and I could imagine that it is not the devel version which is in your PATH? Also make sure to use branch alternative_impl_pr105.

I meant to say that I ran it against my local branch with my patch. I made kscript in my path point to the dev version as well, so I'm confident I'm running the dev version.
The result I get is a project that achieves *both* no red and symlink to the actual script instead of a temp. That is the ideal outcome.

The main problem with resolving things is not the red code (of which you're right, there isn't) but that it doesn't let user modify the actual script, and instead duplicates the code to a temporary one, which defeats the purpose of using --idea in the first place

holgerbrandl:
> The result I get is a project that achieves both no red and symlink to the actual script instead of a temp. That is the ideal outcome.

Indeed that the ideal outcome. Without using includes or preambles this is already the case now. In case of methods or symbols it would indeed be possible to use temp-files for preambles and symlink  these plus the original user script. i've played with such an implementation as well, but ran into problems if the preamble contains import statements. For example in the text processing mode (which in fact is just a built-in custom dsl mode), the preamble is:
```kotlin
 //DEPS com.github.holgerbrandl:kscript-support:1.2.4
import kscript.text.*
val lines = resolveArgFile(args)
```
Putting this into a sym-linked tempfile instead of actually resolving the include will result in red code when running e.g.
```bash
kscript --idea -t "lines.print()"
```
I could imagine import statements in preambles (but potentially also in includes) to be a common use-case. So consequently one might need to resolve not only the preambles but also the includes for `--idea`.

holgerbrandl:
All right, let's do it this way then. I've adjusted the feature branch and merged it into master. I'm closing this PR, thanks a lot for input and the nice discussion.

PR #118, opened 2018-04-07
Improve docopt example

C06A:
This suggest a different output from the example and replaces sample Usage text with one explaining kscript. The usage I put here is different from one in the application. This is how I understood your documentation. If it is incorrect then you may get an idea what in README is not clear.

### DIFF
--- a/examples/classpath_example.kts
+++ b/examples/classpath_example.kts
@@ -13,22 +13,23 @@ import java.util.*
 //val args = listOf("foo", "bar")
 
 val usage ="""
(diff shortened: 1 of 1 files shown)


C06A:
I don't see a reason for these tests to fail. They don't have anything to do with my changes and commands work correctly when I run them locally.

Would you take a look?

holgerbrandl:
Indeed, the example was from some unrelated project. Thanks for the PR. I'll have a look at the tests.

PR #133, opened 2018-06-27
allow to include same file from multiple files

oshai:
add a set of already added files and check if files was already added, in such case do not try to add it again

This happened to me while trying to include same files from multiple files (diamond topology)

got this exception when starting an idea project:

Exception in thread "main" kotlin.io.FileAlreadyExistsException: .../kscripts/utils.kt -> /Users/oshai/.kscript/kscript_tmp_project__ksuite.kts_1530089926297/src/utils.kt: The destination file already exists.
        at kotlin.io.FilesKt__UtilsKt.copyTo(Utils.kt:193)
        at kotlin.io.FilesKt__UtilsKt.copyTo$default(Utils.kt:184)
        at kscript.app.AppHelpersKt.createSymLink(AppHelpers.kt:267)
        at kscript.app.AppHelpersKt.launchIdeaWithKscriptlet(AppHelpers.kt:254)
        at kscript.app.KscriptKt.main(Kscript.kt:140)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.jetbrains.kotlin.runner.AbstractRunner.run(runners.kt:61)
        at org.jetbrains.kotlin.runner.Main.run(Main.kt:109)
        at org.jetbrains.kotlin.runner.Main.main(Main.kt:119)

### DIFF
--- a/src/main/kotlin/kscript/app/AppHelpers.kt
+++ b/src/main/kotlin/kscript/app/AppHelpers.kt
@@ -241,11 +241,15 @@ sourceSets.main.java.srcDirs 'src'
         // https://stackoverflow.com/questions/17926459/creating-a-symbolic-link-with-java
         createSymLink(File(this, scriptFile.name), scriptFile)
 
-
+        val includedFiles = mutableSetOf<String>()
         // also symlink all includes
         includeURLs.forEach {
             val includeFileName = it.toURI().path.split("/").last()
-
+            //make sure not to include each file more than once (it will fail)
+            if (!includedFiles.add(includeFileName)) {
+                return@forEach
+            }
+            
             val includeFile = when {
                 it.protocol == "file" -> File(it.toURI())
                 else -> fetchFromURL(it.toString())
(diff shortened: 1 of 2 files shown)


holgerbrandl:
Thanks for the suggestion. I indeed did not think of diamond topologies here. 

Couldn't we simply do `includeURLs.distinct().forEach {` to avoid the harder to read `return@forEach` and the additional checking logic?

Would it be possible for you to create unit-test example under `kscript/test/resources/diamonds` and to add it to the test-suite? By doing so we could better prevent regressions of this feature.

oshai:
sounds a good idea, I hope I will have time soon to fix

oshai:
I did changes and add diamond file. not sure if this is the testing you meant and where is the suite I should include it.
note I didn't use distinct() but distinctBy{} since it is better to compare strings of the file name itself (or full name, I was not sure if we are creating directories)

holgerbrandl:
Thanks for the PR. `distinctBy` is indeed more elegant here.

holgerbrandl:
The latest release v2.4.5 contains your patch. Thanks again for your help.

PR #135, opened 2018-07-03
add update instructions for mac

oshai:


holgerbrandl:
Thanks @oshai for improving the docs.

Issue #99, opened 2018-02-03
can //INCLUDE  support directory?


howawong:
Currently I can only import files only line by line? Is it possible to include all kt/kts files in a directory?

holgerbrandl:
No this is not possible at the moment. I'd suggest to use a dependency where possible. It's much faster than compiling directories with multiple files along with the script of interest.

Also a fix does not seem straighforward to me, but feel welcome to submit a PR (including impl and tests). See `kscript.app.ResolveIncludesKt#resolveIncludes`

PR #147, opened 2018-08-08
Adds IncludeDir and IncludeDirs directives

cjbrooks12:
This PR will resolve holgerbrandl/kscript#99, adding support for including multiple files in a local directory with a single script directive.

## INCLUDE_DIR

This directive finds all files in the target directory, and replaces the directive with `INCLUDE` directives for each file. This has the effect of including all files in the target directory into the calling script.

**Example**

```
//INCLUDE_DIR path
or
@file:IncludeDir(path)
```

## INCLUDE_DIRS

This directive finds all directories in the target directory, and replaces the directive with `INCLUDE_DIR` directives for each directory. This has the effect of recursively including all files in the target directory into the calling script.

**Example**

```
//INCLUDE_DIRS path
or
@file:IncludeDirs(path)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,8 @@ The following directives supported by `kscript` to configure scripts:
 * `//DEPS` to declare dependencies with gradle-style locators
 * `//KOTLIN_OPTS`  to configure the kotlin/java runtime environment
 * `//INCLUDE` to source kotlin files into the script
+* `//INCLUDE_DIR` to source all kotlin files in a directory into the script
+* `//INCLUDE_DIRS` to recursively source all kotlin files in a directory into the script
 * `//ENTRY` to declare the application entrypoint for kotlin `*.kt` applications
 
 
@@ -255,7 +257,9 @@ println(robustMean)
 ```
 The argument can be an URL, absolute or relative file path. Note that URLs used in include directives are cached locally to speed up processing, that is `kscript` won't fetch URLs again unless the user actively clears the cache with `kscript --clear-cache`.
 
-For more examples see [here](test/resources/includes/include_variations.kts).
+Similarly, you can use `//INCLUDE_DIR` to directly include all files in a single directory, or `//INCLUDE_DIRS` to recursively include all files in a directory and its subdirectories.
+
+For more examples see [INCLUDE here](test/resources/includes/include_variations.kts), [INCLUDE_DIR here](test/resources/include_dir/input.kts), or [INCLUDE_DIRS here](test/resources/include_dirs/input.kts)
 
 
 ### Use `//ENTRY` to run applications with `main` method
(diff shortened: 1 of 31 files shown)


holgerbrandl:
Thanks for PR.

Why do you think that nested Includes are not sufficient? Like you could have one all_utils.kts which includes all the child files. This would still give an actual well defined tree.

Currently `@file:Include` is inspired byhttps://www.rdocumentation.org/packages/base/versions/3.5.1/topics/source and the R devs did not create a recursive or directory mode for a reason I think, which is to keep a rather well defined dependency tree.

Also imho using `@file:Include` is just a temporary workaround before creating an actual jar artifact, which is the better, and more robust solution. So if a user has dozens of includes it would be way faster to jar them and add the jar as a dependency. When using `@file:Include` the complete sources are recompiled whenever you run a changed script.

cjbrooks12:
My main motivation is wanting to use Kscript as the preferred command-line runner for [Orchid](https://github.com/JavaEden/Orchid). Orchid is designed so that it can be bootstrapped and run from multiple different tools for better integration (currently only Gradle, but Maven is on the roadmap). But given that there was a previous issue for directory includes, I figured it would be a good use-case for others and not just myself.

Orchid is a plugin-based tool, and supports loading plugins both as Jars on the classpath, and also private plugins that you want to manage yourself. As a tool for building websites, I want to encourage users to build their own plugins, to customize their site to exactly their needs. When running Orchid with gradle, you have a new `src/orchid` sourceset you can place your files in. Every file in that directory gets compiled and included in the Orchid build, without having to individually specify each file to be used. But when running Orchid from KScript, the only way to include additional files is to list each one individually. It will not allow me to tell users to "drop private plugin code in `src/orchid` to include it in the build", they would have to individually @file:Include each file that they want compiled. 

See [this kscript for reference](https://github.com/copper-leaf/copper-leaf.github.io/blob/master/orchid.kts). It currently includes 3 files, when I really just want to to include everything in the `kotlin/` directory. These files I am including are customized such that they aren't of much use to anyone else unless they want to make this exact site, and so I don't want to go through the trouble of publishing an artifact just for this site. But I would also like the flexibility to just add or remove files as needed to quickly try things out, and to structure the code in an organized fashion so I can easily manage this site as it grows in complexity and customization.

In general, directory includes are a convenience feature for people wanting to just toy around with larger Kotlin projects without needing to set up a full Gradle build, or for creating a better user experience and lower barrier-to-entry for using Kscript as a highly customizable CLI front-end for other tools.

holgerbrandl:
As much as I appreciate your effort @cjbrooks12 , I think it deviates from the original intention too much. Dependencies should mainly be declared with `@file:DependsOn`. `@file:Include` was added a minor extension to this model. Parsing complete directory tree's to reveal a script's dependency model is outside the current scope of `kscript`.

So in your case I'd suggest to deploy the orchid generated scripts as a temporary dependency which can be declared via `DependsOn`.

PR #150, opened 2018-08-17
Replace hardcoded kotlin version to KotlinVersion.CURRENT

hj-lee:
There is no functional change. I think it would be easier if we don't need to change the hardcoded version strings each time we want to change kotlin version.


holgerbrandl:
Thanks for your PR.

Not sure about how other open-source projects handle it, by I actually like the a more defined version indicator, which ties a kscript release always to a well-defined kotlin release. What are the benefits of using KotlinVersion.CURRENT?

oshai:
I also think it is better to be explicit about the kotlin version used.

holgerbrandl:
It's settled then. Thanks for your feedback @oshai.

holgerbrandl:
You're right. I thought you were suggesting to change the version used to build kscript itself.

Issue #49, opened 2017-11-02
get rid of `mvn` and use jar-dependency to resolve //DEPS


holgerbrandl:
E.g. https://www.eclipse.org/proposals/technology.aether/

stephenbyrne-mfj:
Aether seems to be dead. Groovy Grape use Ivy to do it. (https://github.com/groovy/groovy-core/blob/master/src/main/groovy/grape/GrapeIvy.groovy)

holgerbrandl:
Or we could use https://github.com/kohesive/keplin

gildor:
Keplin uses aether under the hood.
Grape also updated last time 3 years ago, later than Aether but still not active development.
Also, even [Gradle uses Aether](https://github.com/gradle/gradle/blob/master/gradle/dependencies.gradle#L155-L160) for maven (at least for publication, as I see, but cannot be sure).

Maybe make sense to ask an advice for someone from Gradle team. But Aether still looks like good option and better than standalone `mvn`

holgerbrandl:
Thanks for the info and the codepointer @gildor.

gildor:
Also, would be good if Aether (or other implementation) could reuse local .m2 cache to prevent dependency downloading only for script

gildor:
@holgerbrandl I've started working on Aether integration in my fork and now have early proof of concept that passes all the tests. I should clean up and probably refactor dependency resolution and tests.

Did you consider something about this feature? Maybe you have some different plans

holgerbrandl:
That's really exciting news! We just have to integrate it carefully to avoid breaking user-installations. There are some dependency tests already, but I'm not sure if they really cover the full spectrum of use-cases yet. 

Only considerations from my end
* maintain option to allow for custom artifact repos (see https://github.com/holgerbrandl/kscript#can-i-use-custom-artifact-repositories)
* maintain artifact download reporting: currently kscript captures some parts of the maven logs and forwards them to the users to indicate progress when downloading dependencies.

Never touched but still on the list are also dependency exclusions https://github.com/holgerbrandl/kscript/issues/12

How much does the shadow jar grow by adding aether?

gildor:
@holgerbrandl 
> maintain option to allow for custom artifact repos
Yes, custom repos already supported

> maintain artifact download reporting
I have some logs already, no progress tho, just message about artifact downloading

> dependency exclusions
Shouldn't be a problem, Aether supports exclusion

> How much does the shadow jar grow by adding aether
[...]

Issue #154, opened 2018-09-01
It fails to call bash internally on windows when using cygwin


deviant-studio:
Hi guys! 
Do you have plans to make kscript windows compatible?

holgerbrandl:
It should run under cygwin. At least we implemented a few tweaks around it
a while ago. Please submit issues as you encounter them

I have not tried the Linux subsystem yet, but since it runs with bash,
kotlin and a jdk, kscript should run as well.


Mykola D <notifications@github.com> schrieb am Sa., 1. Sep. 2018, 13:08:

zaxebo1:
probably someday **this part https://github.com/holgerbrandl/kscript/blob/master/src/kscript which has been written in bash shell , can be ported to powershell "also"** . That will facilitate it to run with powershell,kotlin and jdk too on MSWindows. ie, no need of cygwin then. Just drop it on windows and run.

holgerbrandl:
I guess deployment via sdkman wont work. But if you could provide a powershell launcher via PR, I could add at least a windows release format for download. What do you think?

We could even think about testing it on travisCI with the mentioned linux-support for powershell.

abelsromero:
Installation with sdkman works fine but it just gets stuck and never executes. Is there some verbose more to get info or some way to debug it?

Btw, on WSL works fine.

holgerbrandl:
Could you share a specific example? Just to make sure that it runs fine under linux/macos.

Also please note that a few windows related issues were fixed recently via #175 but are not yet released.

abelsromero:
The one I tested was just this:

```
#!/usr/bin/env kscript
import com.github.kittinunf.fuel.httpGet

//DEPS com.github.kittinunf.fuel:fuel:1.15.1

fun getToken(): String {
    val (request, response, result) = "https://httpbin.org/get".httpGet()
            .responseString()
    return result.get()
}
println(getToken())

```
But a Hello world fails as well, in this case, the "Resolving ...." is not shown, so I guess the script is partially processed.

```
#!/usr/bin/env kscript
println("Hello from a Kotlin script!")
```

> Also please note that a few windows related issues were fixed recently via #175 but are not yet released.

I'll give a try with the master version and come back.
Thanks!

abelsromero:
Same issue with a jar build from master

holgerbrandl:
I've tried your example and it runs fine with macos/linux at least.

Could you check if `kscript --clear-cache` might solve the problem?

[...]

Issue #143, opened 2018-07-23
feature: Add credentials to @file:MavenRepository


MikeMitterer:
I'm using Artifactory for my Maven-Repo
In my case the repo needs credentials

Would be cool if I could specify them for @file:MavenRepository

holgerbrandl:
Is there a public maven repo with credentials support which we could use for testing?

MikeMitterer:
Not that I know - can I test something for you?

stephenbyrne-mfj:
Hard-coding credentials in a script is probably a Bad Idea. You can put your credentials in `~/.m2/settings.xml`:

    <?xml version="1.0"?>
    <settings>
        <servers>
            <server>
                <id>{the ID from the @file:MavenRepository}</id>
                <username>azurediamond</username>
                <password>hunter2</password>
            </server>
        </servers>
    </settings>

This works today, I just tested it on master.

holgerbrandl:
I'd also think that credentials should better go into config file as described above, but it would be a technically easy to extend the annotation to accept credentials. Doing so would also help to see if  #49 could provide a better solution for dependency resolution.

@ligee Are you planning to add credential support to `@file:Repository` as used in https://github.com/Kotlin/KEEP/blob/master/proposals/scripting-support.md#script-with-maven-dependencies-resolving-support ?

PR #159, opened 2018-09-07
#49 use jcabi-aether for dependency resolving

andreas-mausch:
Sample output:
```
$ kscript /home/neonew/Schreibtisch/kscript/test/resources/depends_on_with_type.kts
[kscript] Resolving dependencies...
[kscript]     Resolving org.javamoney:moneta:1.3@pom...Done
[kscript]     Resolving com.github.holgerbrandl:kscript-annotations:1.2...Done
[kscript] Dependencies resolved
getBigDecimal(1L): 1
```

- Updated shadow plugin to 2.0.4 (otherwise the build is extremely slow)
- Resulting kscript.jar went from 122kb -> 18.5mb :(
- I've excluded some (unused?) dependencies to strip it down to 5.1mb
- No requirement for mvn anymore

Some excludes were even needed, otherwise we would see:
Sep 07, 2018 4:03:24 AM org.hibernate.validator.internal.util.Version <clinit>
INFO: HV000001: Hibernate Validator 5.1.2.Final
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.

Feedback is welcomed.


holgerbrandl:
Great PR, thanks for this lovely piece of work.

Will it reuse the local .m2 cache as commented on by @gildor in #49?

Some (@stephenbyrne-mfj) commented that aether is no longer under development. So hopefully we're not betting on the wrong horse here. 

I am mostly worried about backward compatibility but it's hard/tedious to test it. So I'll just test the PR locally for a few days, before merging it into master.

andreas-mausch:
> Will it reuse the local .m2 cache

Well, that's a bit of a problem. Currently I do:

```
File(System.getProperty("user.home") + "/.m2/repository")
```

So M2_HOME can be different than ~/.m2 and I currently ignore any settings.xml which overwrites the localRepo location.
Won't be a problem for most users though.

> aether is no longer under development

True. Gildor mentioned "Also, even Gradle uses Aether for maven (at least for publication, as I see, but cannot be sure)."
But absolutely, this is a risk.
I've addtionally used jcabi-aether as a wrapper, which has a couple of commits other the last year, but not sure how well it is maintained overall.

> I am mostly worried about backward compatibility but it's hard/tedious to test it.

I have the same doubts, and I won't be disappointed if you decide to stick with the maven solution.
This PR doesn't win much (maybe a bit ramp-up time because we can skip pom parsing) and is a bit risky.

I just wanted to share it and hey, maybe it's working well.
If you decide you want to merge it, I will definitely check it doesn't break the dependencies.txt.

dimiii:
@andreas-mausch Could you add some tests, please?

andreas-mausch:
@dimiii Do you have a special test in mind?

I see this more as a refactoring. All scripts in the test folder use the new method, so as long as they are green, it should be fine.

holgerbrandl:
I agree with @andreas-mausch , as long as the tests are green the PR is ready for merging.

Sorry for the delayed process. But since kscript maybe merged into the offical kotlin-core soon, we may want/need to use the official scripting aether integration instead.

abelsromero:
I tested on mingw (gtibash) and resolution works, execution still fails though, but it is a great step forward.

But the Kotlin version needs to be updated to `1.2.70` (same as current master) to avoid `e: Kotlin home does not exist or is not a directory: `.

holgerbrandl:
This was to be expected because of the same issue as described in #154: Before calling `kotlinc` we check its presence by calling bash, which fails under windows at the moment.

abelsromero:
On thing at a time, what is needed to merge this?

holgerbrandl:
Not much, it's a great PR. There's two arguments

1. It's not adding anything new from a user-perspective, and why change a running system? 
2. With @ligee's effort to port `kscript` to the new Kotlin scripting API (since v1.3), I guess the PR would become obsolete, because the scripting API has its own aether integration for dependency resolution. 

Clearly `kscript` failing on windows somehow invalidates (1) but the PR would imho not fix this entirely because execution would still fail on windows.

abelsromero:
>     2\. With @ligee's effort to port `kscript` to the new Kotlin scripting API (since v1.3), I guess the PR would become obsolete

Would not the whole project become obsolete?

holgerbrandl:
The idea is rather to make it part of the core distribution of Kotlin :-). 

So technically this repo would become obsolete, but conceptually it would imho be a great grand finale for the idea of providing  an _easy-to-use, very flexible, and almost zero-overhead solution to write self-contained mini-applications with Kotlin_. (quoting myself from the readme)

it's an ongoing process so let's see if we manage. Until then I'll keep working on it.

abelsromero:
Glad to hear that! I'd like to give it a try to fix the issues to make it run with Windows.
And it will be much easier if this PR gets merged.

abelsromero:
I have fixed most of Windows issues (https://github.com/abelsromero/kscript/tree/add_windows_support) but I am finding problems with the current maven resolution due to issues parsing the output of the mvn command from cygwin (line ending + pathing stuff). Honestly, I don't see the point in fixing those having this PR which makes the resolution much more stable and clean.

So, apologies if I sound rude, I just want to be 100% clear here, so I ask _can this PR be merged?_. As soon as this is merged I can rebase and test it with my fixes. Note I am also testing on Mac and Linux, and checking with `local_test_suite.sh`.

Answering the arguments given before:

> 1. It's not adding anything new from a user-perspective, and why change a running system?
> 2.  With @ligee's effort to port `kscript` to the new Kotlin scripting API (since v1.3), I guess the PR would become obsolete, because the scripting API has its own aether integration for dependency resolution.

1. This is clearly a big improvement and removes depending on maven.
2. So long there's no clear date on Kotlin 1.3 release, kscript is the only way to provide kotlin scripting. Not to mention this can be a tool for kotlin <1.3, I don't think this is going to become obsolete.

abelsromero:
Apologies again, I admit I was a bit heated yesterday and overreacted.
However, I maintain my request so we can provide Windows compat.

holgerbrandl:
I agree waiting for v1.3 or some future conversion/migration is not really helpful. So let's do it then. I've cleaned most open tickets this morning and pushed out a new release, so that we can focus on this PR now.

We need to 

1. Fix the conflicts in the PR with master
2. Add some tests for https://github.com/holgerbrandl/kscript/issues/166 and https://github.com/holgerbrandl/kscript/issues/143. Both relate to reoccurring reported use-cases so we should make sure that aether would support them in some sensible way. They are not supported yet, but could be adressed by moving the dependency resolution from `mvn` to `gradle`, so I'd love to ensure that `aether` allows for the same functionality.
3. Merge it. :-)

I think your windows-compat improvements (thanks!) should be done separately (like in a different PR/branch). Not sure if we should merge before or after this one. What do you think? We also need to think about testability: Is there any way we can test windows compat on Travis or some other CI platform that integrates with github?

abelsromero:
Thanks for you comprehension.

> I think your windows-compat improvements (thanks!) should be done separately (like in a different PR/branch). Not sure if we should merge before or after this one. What do you think?

I agree it should be a separate PR, but after this one, I found issues with running mvn command that honestly I had not clue. For instance, some poms resolved correctly, others not, but only form shell, running from a junit they resolved correctly.
But this raises the question @andreas-mausch, can you take care of the extra tests and conflicts?

> Is there any way we can test windows compat on Travis or some other CI platform that integrates with github?

Travis has just recently added windows, or else we can integrate AppVeyor. I have used AppVeyor in many project before, shouldn't be hard to begin with. I can add that to the windows PR, but you'll have to enable it for your repo.

andreas-mausch:
@abelsromero Hi, luckily I'm on vacation for the next weeks, so I should find some time to take care of the conflicts.
For the additional features: I need to look into them, but from my first impression they seem to be simple.

From my memory: There was also an issue to support a version format like the "1.2+" notation, like gradle supports it. Imo this will require some more effort and I wouldn't implement it (right now). I also really dislike non-fixed versions, so it's also a matter of my personal choice here. ;) Just want to mention it because of "moving the dependency resolution from mvn to gradle".

andreas-mausch:
Conflicts in the PR with master: Resolved.

holgerbrandl:
Thanks @andreas-mausch for this great PR and sorry again for the endless delay on my end.

abelsromero:
> sorry again for the endless delay on my end.

I missed that too 🤦‍♂️

andreas-mausch:
Np, glad you like it. Thanks for merging.

Issue #162, opened 2018-09-10
run spring boot with kscript


jangalinski:
I tried running

```
#!/usr/bin/env kscript

@file:DependsOn("org.springframework.boot:spring-boot-starter:2.0.4.RELEASE")

package my.package

import DependsOn
import org.springframework.boot.autoconfigure.SpringBootApplication
import org.springframework.boot.SpringApplication

@SpringBootApplication
class DummyApp

SpringApplication.run(DummyApp::class.java, *args)
```

it starts but fails with

```
ERROR org.springframework.boot.SpringApplication - Application run failed
java.lang.IllegalArgumentException: No auto configuration classes found in META-INF/spring.factories. If you are using a custom packaging, make sure that file is correct.
	at org.springframework.util.Assert.notEmpty(Assert.java:450)
```

A comparable groovy/@Grab script with the same setup works, so I wonder if this is something that can/will never work with kscript due to design or is just a missing feature that can be achieved ...  somehow

I know that running complex spring boot apps with kscript seems to be out of scope, but I have quite a experimental microservices that require minimal coding and would benefit from a scripting approach ...

holgerbrandl:
I actually love the use-case.

It's not working for me either. It's because we don't use spring's custom packaging method.

The problem seems to center around the missing `META-INF/spring.factories`. However, even if I patch the jar created by kscript internally to include the file, it fails with the same error. 
So to progress here, we'd need to find  a way to provide `META-INF/spring.factories` via classpath. It seems partially documented under https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-developing-auto-configuration.html#boot-features-developing-auto-configuration
In particular is says
> Auto-configuration classes can be bundled in external jars and still be picked-up by Spring Boot.

So if we could find a way to run a spring-app without their deployment-scheme by providing an extra-jar that adds the missing `META-INF/spring.factories` the classpath, we could configure `kscript` to run it accordingly.

jangalinski:
glad you like the idea ... unfortunately, I haven't looked that deep into how kscript is working (just was glad that it IS working :-), so I won't be much help until I do ... but I can do research and testing if required.

I found this https://stackoverflow.com/questions/38792031/springboot-making-jar-files-no-auto-configuration-classes-found-in-meta-inf quite useful ... but it more or less resembles what you already found.

holgerbrandl:
[...]

PR #165, opened 2018-09-14
update readme.md

CuriousNikhil:
add link to a blog post


holgerbrandl:
Great article. Its awesome to read that we can run a complete webserver with ktor via kscript. Unfortunately this does not seem possible with spring boot due to the way it relies on custom jar-packaging. See #162.

CuriousNikhil:
Right. I think Ktor is more lightweight thank spring-boot. i'll try using Coroutines feature with this backend thing.

PR #169, opened 2018-09-26
fix maven open version ranges

### LINKED ISSUE #166: dependency notation does not support maven version range syntax containing `)` and `,`
EDIT: Maven supports version ranges natively: examples: http://www.mojohaus.org/versions-maven-plugin/examples/resolve-ranges.html
an it works in kscript once the annotation and comment parsing is not messing up on `,` and `)`

old issue content:

i am writing a library and tool that i try to make compatilble / consumable by kscript
the versions are pushed to maven by jenkins , each git commit increments the build number

ideally i would like to use
```
@file:DependsOnMaven("moe.nikky.voodoo:dsl:0.4.0+")
```
instead of
```
@file:DependsOnMaven("moe.nikky.voodoo:dsl:0.4.0-132")
```

NikkyAI:
specifying `[1.0)` as version was broken, because kscript would not pass the correct values to the POM

disallow `,` in DEPS_COMMENT
`;` and space should still work

improve parsing of DEPS_ANNOT
use `.substringBeforeLast(")")`instead of `.split(")")[0]`
assume there is no more `)` after the annotation
maybe also make sure `//.+` is stripped before?

is there no smarter annotation parser available ?

fixes #166

### DIFF
--- a/src/main/kotlin/kscript/app/Script.kt
+++ b/src/main/kotlin/kscript/app/Script.kt
@@ -97,7 +97,7 @@ private fun extractEntryPoint(line: String) = when {
     line.contains(ENTRY_ANNOT_PREFIX) ->
         line
             .replaceFirst(ENTRY_ANNOT_PREFIX, "")
-            .split(")")[0].trim(' ', '"')
+            .substringBeforeLast(")").trim(' ', '"')
     line.startsWith(ENTRY_COMMENT_PREFIX) ->
         line.split("[ ]+".toRegex()).last()
     else ->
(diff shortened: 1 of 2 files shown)


NikkyAI:
it has become obvious to me this attempt is fighting windmills,
the range notation would have to be converted to gradle (which has less finegraned control, so it would be lossy)
also we cannot fix the `//DEPS` comment in fear of breaking old code

i will open a new PR 
where i will instead convert gradle's `+` (and pretty much only that) to maven ranges
since  supporting the full ranges syntax is impossible (without breaking `//DEPS`) and this covers most the usecases already

PR #170, opened 2018-09-26
support dynamic versions ending with `+` 

### LINKED ISSUE #166: dependency notation does not support maven version range syntax containing `)` and `,`

NikkyAI:
for maven these are wrappend in a range
`1.2.3+` -> `[1.2.3,)`

for gradle they are not touched and just passed as is

closes #166

### DIFF
--- a/src/main/kotlin/kscript/app/DependencyUtil.kt
+++ b/src/main/kotlin/kscript/app/DependencyUtil.kt
@@ -117,11 +117,20 @@ internal fun buildPom(depIds: List<String>, customRepos: List<MavenRepo>): Strin
             quit(1)
         }
 
+        // replace + with open version range for maven
+        val version = matchResult.groupValues[3].let {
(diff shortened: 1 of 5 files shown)


holgerbrandl:
Thanks @NikkyAI for this beautiful PR, especially since it even contained unit and shell integration tests.

PR #175, opened 2018-10-11
cygwin support improvements

vehovsky:
Currently kscript is not working on Windows in cygwin.

> /cygdrive/c/kotlin/scripting
$ kscript 'println("hello world")'
error: please specify at least one name or file to run

Most of the issues were caused by not quoting paths containing windows path separator '\' that was actually escaping characters in cygwin.

The '--package' didn't work for similar reason. In the generated gradle build file the actual script was added with windows path separators '\'. Now using invariant separator '/'.

### DIFF
--- a/src/main/kotlin/kscript/app/AppHelpers.kt
+++ b/src/main/kotlin/kscript/app/AppHelpers.kt
@@ -244,7 +244,7 @@ sourceSets.main.java.srcDirs 'src'
         // also symlink all includes
         includeURLs.distinctBy { it.fileName() }
           .forEach {
-            
+
             val includeFile = when {
                 it.protocol == "file" -> File(it.toURI())
                 else -> fetchFromURL(it.toString())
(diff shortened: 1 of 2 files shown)


holgerbrandl:
Looks like a great PR, thanks for contributing. Could you merge-in my typo fix of the build-file, to make sure that the tests are still fine.

vehovsky:
Hi @holgerbrandl sure it's done. 

3 tests are failing. But those tests are failing for me on master locally even without any changes.

Could you have a look at those 3 tests? 

Thanks

holgerbrandl:
I've adjusted the tests to cope with the additional quoting in a separate branch pr175 (see link above in this thread).

I've also merged in PR although I realized afterwards, that it seems to break the packaging by not setting the executable flag on the packaged script. See https://travis-ci.org/holgerbrandl/kscript/builds/442050672
I like the simplified usage of `reallyExecutable` but we may need to set the `+x` manually now or roll it back I guess.

vehovsky:
You are right, I also noticed yesterday, but forgot to fix it. Can you please just change

> AppHelpers.kt

`File(tmpProjectDir, "build/libs/${appName}").copyTo(pckgedJar, true)`

to:

`File(tmpProjectDir, "build/libs/${appName}").copyTo(pckgedJar, true).setExecutable(true)`

holgerbrandl:
All tests are green again. :-) Thanks for contributing to kscript!

PR #182, opened 2018-10-21
Add a basic template example to examples folder

ivmos:
Just in case you want to add https://gist.github.com/ivmos/af857cb431cb73d2a7a86acc71785616 to examples folder. It's a template I use for my little kscripts.

BTW congratulations for this great project and happy Hacktoberfest :)

### DIFF
--- a/examples/basic_template_example.kts
+++ b/examples/basic_template_example.kts
@@ -1,0 +1,27 @@
+#!/usr/bin/env kscript
+// TODO fill here what your script does
+
+import kotlin.system.exitProcess
+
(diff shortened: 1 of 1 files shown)


holgerbrandl:
Hi, thanks for contributing to kscript. Though I'm not sure if your example adds much on  top of the existing ones like `classpath_example.kts`. What do you think? Whats missing in the existing examples?

ivmos:
I think the (minor) benefit of this other one is that it works out of the box with the flow of doing kscript --idea script, whereas classpath_example.kts seems not (at least for me I see I would need to add Docopt to build.gradle).

I know it's not a big deal but I found more convenient avoiding that. Same thing with colors for console, I tried a few libraries to achieve it and I found easier just adding those support lines.

I wanted a straightforward basic script where I can open with kscript --idea, fill the stuff I need and ready to go, without needing to bother even to make the template work every time.

That said, feel free to avoid the merge, I understand this can be very personal.

Thanks :)

captura de pantalla 2018-10-22 a las 19 37 55

holgerbrandl:
The `classpath_example.kts` fails because of https://youtrack.jetbrains.com/issue/KT-24789 which is a kotlin platform issue, so you may want to vote for it. It's indeed an annoying bug.

So in this regard you are right that having a dependency-free example is useful at the moment.

PR #185, opened 2018-10-26
Support "~" in INCLUDE

hj-lee:
Hi,

This change enables to include files relative to home directory like below:

@file:Include("~/my-kscript/utils.kt")


holgerbrandl:
Thanks for contributing to `kscript`.

PR #186, opened 2018-10-28
Don't hardcode the KSCRIPT_HOME path in Travis

egor-n:
This makes it so tests can pass on forks too if contributors enable them in Travis.


holgerbrandl:
Thanks for contributing. I never noticed this selfishness in the CI config.

PR #187, opened 2018-10-28
Improve file-type detection when fetching from URL

egor-n:
Just adding a failed test case for now.


egor-n:
Is there a more reliable but still simple way to check for the presence of `fun main`?

holgerbrandl:
Great PR. Thanks. 
It's a fallback heuristic, so I guess scanning for `fun main` is just fine here until somebody comes up with a counter-example.  (which is for sure easy to construct but unlikely to happen in real life)

Issue #201, opened 2018-12-06
`kscript --idea scriptWithoutExtension` copies a script instead of symlinking it


EugeneSusla:
Scripts without .kt[s] extension already run fine with kscript, with the only remaining broken piece being the `--idea` experience.

So I think it would be better to symlink the script instead of copying it, adding .kt[s] extension so that idea correctly recognizes it.
Simplest way would be to just always symlink as .kts, although a more sophisticated implementation might try to figure out if the file was meant as .kt instead.

Motivation for extension-less scripts support/usage:
With #195 now merged, scripts can now be shared with kotlin/kscript being a hidden implementation detail, that the end user doesn't have to be educated about, which is way less friction.
The only remaining hurdle remains that the end user likely have never dealt with .kt[s] files before, so might not know it can be just run directly.
It is also more conventional for command-line tools to not have an extension, e.g. `kscript` instead of `kscript.sh`

PS: Happy to work on a PR if this looks ok. I believe there should be no downside to changing this behavior.

holgerbrandl:
> scripts can now be shared with kotlin/kscript being a hidden implementation detail, that the end user doesn't have to be educated about,

+1

kscript also supports regular kotlin `kt` files https://github.com/holgerbrandl/kscript#use-entry-to-run-applications-with-main-method which might interfere with a quick fix.

So, to implement your suggestion the/an easy way would be to replace usages `scriptFile.extension` with an extension method that infers the type if needed from the content if an extension is missing. Like 
```
fun File.guessExtension(): String {
    return if (extension.isNotBlank()) {
        extension
    } else {
        if (Script(this).findEntryPoint() != null) "kt" else "kts"
    }
}
```

Then we'd need another else in 

https://github.com/holgerbrandl/kscript/blob/4dabdd729f5c0706ab04343dc6a0b0dd77d27bf7/src/main/kotlin/kscript/app/Kscript.kt#L363

to check if the argument file is a kscript without an extension (e.g. by checking the shebang) and avoiding the creation of a temporary kts (as it down at the moment in the final `else`). We may need to be careful here, because it may be that e.g. file substitution handles (which also come without an extension) can be just read once.

So in total I'm not yet sure if the benefits (it's a minor feature only) outweigh the putative complexity of the patch. However, if you're still interested and would manage to put it into a reasonally compact PR (including a unit test if possible) feel welcome.

holgerbrandl:
Rejected PR, so closing this ticket as well

PR #202, opened 2018-12-12
Support `kscript --idea scriptWithoutExtension`

EugeneSusla:
This implements #201

(also automated a few steps for running `test_test/suite.sh`)

Test plan: `test/test_suite.sh`, as well as manually reviewing results of `kscript --idea`


holgerbrandl:
Thanks for your contribution.

1. My biggest concern is about the changes in  `prepareScript()`. Do we really want/need to symlink at this level? This is risky since there might be other scripts with the same name, and the symlink might even exist already. I thought that we could simply use the existing file without symlink. I'd think that we would just need to use a better `extension` method such as the suggested `guessExtension()` in #201 to patch `prepareScript` and https://github.com/holgerbrandl/kscript/blob/e4c34d60b068d3f25b656e4eaa0f4fbbf576b14b/src/main/kotlin/kscript/app/Kscript.kt#L164

2. `Script.hasMainMethod` should also support parameterless mains introduced with kotlin v1.3, see https://kotlinlang.org/docs/reference/whatsnew13.html#parameterless-main

Concerning the test-changes (which could have gone into a separate PR for simplicity and ease of discussion):

1. With the changes to `TestReadme.md` a separate `kscript` installation is now required to follow the instructions. Without `kscript --clear-cache`. 

2. I don't really like `export KSCRIPT_HOME="$(dirname $0)/.."`. Before the tests could be called from anywhere as long as KSCRIPT_HOME was defined, but no we make the silent assumption that the user calls them from the working copy root. 

3. Using a mock idea is a good idea, but we should remove the duplicate mock from the travis config, and make push the PATH adjustment to the top of the test-suite because the mock is used by other tests as well (e.g. see `kscript_nocall --idea test/resources/includes/include_variations.kts `)

4. Running the tests in `/tmp` is a good idea, but not all tests are ready for this because they use relative links to resolve resources. (e.g. see `--package test/resources/package_example.kts`)

EugeneSusla:
Sorry for bundling the test changes - they were an afterthought tbh

1. I tried going that way at first, but learned that kotlin compiler requires input files to be .kt[s], so can't use original file directly. The original approach of copying as .kts turns out to also not would have worked for implicit .kt file.
IIRC temp dir where I symlink it is unique per kscript invocation, so collision should only occur if there's multiple distinct scripts with same nameWithoutExtension that include each other. This is already mostly unsupported by `--idea` anyway, so I'm hoping I'm not introducing any noteworthy new corner case with this.
2. Good catch! Had a feeling I might be missing something like that

Re: test
1. Not sure I understand that - afaict requirements for neither separate installation nor `--clear-cache` shouldn't have changed with the instructions update. Possibly related - I should probably also replace `kscript` with `./build/libs/kscript` in `TestReadme.md`
2. Oops, didn't realize this'll end up making KSCRIPT_HOME relative - was intending to keep your use-case. Will fix to make it work from anywhere
3. Will do
4. Hmm, don't know why I assumed they would. Will check how easy it is to adapt, and undo the /tmp part otherwise

holgerbrandl:
Your changes to the PR look good to me although some aspects (remove travis mock, parameterless main, /tmp running) seem to be still open.

re1: that's an annoying limitation of the kotlin compiler. I also did not know about it yet.

> collision should only occur if there's multiple distinct scripts with same nameWithoutExtension that include each other

Indeed, but this makes `--idea` less applicate (which was the main intent of this PR): With the current approach, the user can reopen a project created with `kscript --idea` later. With tmp-symlinks this won't work anymore. I'm not sure about you feel it, but I'd prefer to symlink into KSCRIPT_CACHE_DIR using maybe the scripts path to build a  `extless_<hashcodeofpath>.(kts|kt)` extension. 

re test-1: Before kscript was added to the PATH via KSCRIPT_HOME before clearing the cache. Now since the PATH is changed in the test-suite.sh, it may be missing (unless the user has another globally installed kscript).

---
I've spotted another aspect, which we should clarify:

Why would we need to change the symlinking for _includes_ in https://github.com/holgerbrandl/kscript/blob/f202b8c18eaa36e7c7c264eafa7fa9ef6f4bb4b4/src/main/kotlin/kscript/app/AppHelpers.kt#L268 ? They should never come without an extension. 
Unless there's a reason, I'd suggest to drop this particular change.

EugeneSusla:
Latest push should address the comments.
I went with a slightly more involved solution for the name collision problem to keep `--idea` symlinks suffix-free for the case without name collisions

EugeneSusla:
@holgerbrandl, is there anything else you'd like me to address here?

holgerbrandl:
No @EugeneSusla, I've recently started a new job, and will need some  more days to adjust, before coming back to kscript and working through the growing list of open PRs. Sorry for the delay.

holgerbrandl:
I won't include this, because it's adding too much code complexity for a sine supposedly rarely used functionality. A simple workaround would be
```
ln -s foo foo.kts
kscript --idea foo.kts
```

Sorry for inconvenience.

PR #203, opened 2018-12-13
Fix cache check bug on Windows

hj-lee:
This is a simple bug fix PR. After aether change, kscript does dependency resolution repeatedly on Windows.


holgerbrandl:
Thanks @hj-lee for the fix.

PR #205, opened 2018-12-23
Reuse JVM

hj-lee:
As discussed on #102, this PR processes KOTLIN_OPTS in kscript bash.

The kotlin code differences between #102 and this are:
- It is one commit 
- KOTLIN_OPTS processing code has been removed, and it is lazy operation for --package

Now, scripts using KOTLIN_OPTS can enjoy the startup time improvement by reusing the JVM at the expense of complex kscript bash script.


holgerbrandl:
Closed because of inactivity.

Issue #197, opened 2018-12-01
`--package` not working with Gradle 5.0


abelsromero:
Running the package option with gradle 5.0 installed causes it to fail with the underlying gradle message.
```
Gradle does not allow passing null for the configuration action for CopySpec.from()
```
Note that the message was captured debugging.

With gradle 4.10.2, it worked fine.

abelsromero:
Seems to be a known issue with no quick solution https://github.com/danthegoodman/gradle-capsule-plugin/issues/38

holgerbrandl:
Thanks @abelsromero  for the pointer. So let's hope that they fix it soon. Not sure what to do about it on the kscript side.

abelsromero:
> Not sure what to do about it on the kscript side.

To keep current implementation we could "reverse engineer" the  gradle plugin and directly use the capsule API.
Or just build it the package in another way, I guess that so long as we add a bootstrap script to the path and have the jars somewhere we keep the feature the same from a user stand point. It would be nice also because that way we remove the dependency from gradle.

holgerbrandl:
I think the best approach would be to figure out and implement a PR for the `gradle-capsule-plugin` which fixes the problem.

f-meloni:
Apparently there is already a PR for that
https://github.com/danthegoodman/gradle-capsule-plugin/pull/39
But the problem is no longer really maintained, then could be worth evaluate remove it as dependency to avoid this kind of problems in the future.

holgerbrandl:
danthegoodman/gradle-capsule-plugin#39 seems to work only with gradle 5 and not with older gradles, otherwise I'd just fork the PR and push it to jcenter as a quickfix. But this option does not seems on the table, because at least gradle 4 should be supported by the plugin.

For sure the best solution would be to implement packaging in-process without calling an external tool. This would also remove the dependency on the capsule-gradle-plugin. Maybe capsule could be called programmatically? I would be very thankful for any help on such a feature. 

Alternatively with the new bootstrap header support contributed with #199 we could also think about dropping `--packaging` entirely? IMHO both are a bit too different, so it may be nice to maintain both packaging modes separately.

gianluz:
we have 2 new PRs https://github.com/danthegoodman/gradle-capsule-plugin/pull/41
and https://github.com/danthegoodman/gradle-capsule-plugin/pull/40

holgerbrandl:
[...]

PR #210, opened 2019-02-04
Updated gradle capsule plugin for gradle 5.0 compatibility

gianluz:
https://github.com/holgerbrandl/kscript/issues/197

I have forked the current capsule plugin that was not maintained, hopefully works with this

@holgerbrandl could you double check please if now is working?

Thanks


holgerbrandl:
Thanks @gianluz for this great fix.

PR #211, opened 2019-02-05
Fix package path on text processing mode doc

Arch-vile:
I believe it should be `kscript.text.resolveArgFile(args)` instead of `kscript.text.StreamUtilKt.resolveArgFile(args)` ?


holgerbrandl:
Thanks @Arch-vile for the fix.

PR #219, opened 2019-03-12
Make INCLUDE directive files relative to script dir

jhby1:
The documentation is a bit unclear on what the INCLUDE directive paths are supposed to be relative to - the workdir or script dir? This change makes them relative to the script dir, so that scripts can be invoked from any workdir.

**Old behavior:**

Main script file: "my-script":
```
#!/usr/bin/env kscript

//INCLUDE src/helpers.kt
doIt();
```
Support file: "src/helpers.kt":

```
fun doIt() {
    println("Do something")
}
```

This works: `$ ./my-script`
This fails: `$ cd src/; ../my-script`
[kscript] [ERROR] Failed to resolve //INCLUDE 'src/helpers.kt'

**New behavior:**
This works: `$ ./my-script`
This works: `$ cd src/; ../my-script`

### DIFF
--- a/src/main/kotlin/kscript/app/Kscript.kt
+++ b/src/main/kotlin/kscript/app/Kscript.kt
@@ -333,6 +333,7 @@ fun prepareScript(scriptResource: String): Pair<File, URI> {
             // if we can "just" read from script resource create tmp file
             // i.e. script input is process substitution file handle
             // not FileInputStream(this).bufferedReader().use{ readText()} does not work nor does this.readText
+            includeContext = this.absoluteFile.parentFile.toURI()
             createTmpScript(FileInputStream(this).bufferedReader().readText())
         }
     }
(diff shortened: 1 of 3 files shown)


holgerbrandl:
Thanks for the PR. I am/was under the impression that we resolve includes relative to the script. See  https://github.com/holgerbrandl/kscript/blob/2257bd836b2806733c958a8111bda280f2bc9cc3/test/resources/includes/include_variations.kts for one of our passing unit-tests.

I've tested the existing master (without the PR) using the example `include_variations.kts` but in interpreter mode (ie by prefixing the example with `#!/usr/bin/env kscript`. It works as expected. Both when triggered with absolute path and also with ` ../include_variations_interpreter.kts`. 

Could you provide test which fails with the current master?

holgerbrandl:
Thanks for the update. I'll have a look asap

PR #229, opened 2019-06-24
Add support for trailing logic after @file:Include

rcaught:
This allows for running logic after the file include annotation.

For example: `kscript '@file:Include("file.kt"); myFunction()'`.

### DIFF
--- a/src/main/kotlin/kscript/app/ResolveIncludes.kt
+++ b/src/main/kotlin/kscript/app/ResolveIncludes.kt
@@ -77,7 +77,7 @@ internal fun extractIncludeTarget(incDirective: String): Pair<String, String?> {
         incDirective.startsWith(INCLUDE_ANNOT_PREFIX) -> {
             with(incDirective.replaceFirst(INCLUDE_ANNOT_PREFIX, "").split(")", ");", limit = 2)) {
                 // TODO: Once on Kotlin 1.3, the following conditional can become `this[1].ifBlank { null }`
-                Pair(this[0].trim(' ', '"'), if (this[1].isNotBlank()) this[1] else null)
+                Pair(this[0].trim(' ', '"'), this[1].takeIf { it.isNotBlank() })
             }
         }
         // Comment directives cannot have trailing logic (as it would still be within a comment)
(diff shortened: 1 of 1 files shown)


holgerbrandl:
I'm not yet in favor of this PR, because

1) it is not a generic enhancement for annotation processing and just applies to `@file:Include`. Since we don't properly parse scripts but rather use regexes to parse out configuration from annotations, I think its cleaner to assume config annotations to be present one at time/line.

Allowing for multiple statements opens the box of Pandora, e.g. what about `@file:Include("foo.kt"); @file:Include("bar.kt")`?

2) I simply can't imagine a use case for this new feature. Could you provide some examples?

rcaught:
> I'm not yet in favor of this PR, because
> 
> 1. it is not a generic enhancement for annotation processing and just applies to `@file:Include`. Since we don't properly parse scripts but rather use regexes to parse out configuration from annotations, I think its cleaner to assume config annotations to be present one at time/line.

Understand if that's the case.

> 
> Allowing for multiple statements opens the box of Pandora, e.g. what about `@file:Include("foo.kt"); @file:Include("bar.kt")`?

Yeah, not great.

> 
> 2. I simply can't imagine a use case for this new feature. Could you provide some examples?

I _can_ run what I'm doing as:
```
$ kscript '@file:Include("file.kt")
myFunction()'
```

But I wanted to not have the bash newline, hence this PR.

I'm intending to replace some bash scripts in AWS CodeBuild with Kotlin / kscript.  With [CodeBuild you have phases and sequences](https://docs.aws.amazon.com/codebuild/latest/userguide/build-spec-ref.html#build-spec-ref-syntax), so for each phase and sequence, I just want to include my common file (that has other includes and stores common logic) and then call a method specifically written for that phase and sequence.  No `main` methods, just plain invocation.

Feel free to close if you think it's best not to follow this path. 😄

holgerbrandl:
We will keep the PR open to allow other users to share their opinion as well. As detailed out above I'm not convinced that the feature increment outweighs the additional complexity in parsing/inclusion logic 

For the meantime, a workaround might be to use a custom interpreter for your aws usecase. See https://github.com/holgerbrandl/kscript/blob/master/docs/user_guide.md#create-interpreters-for-custom-dsls for details and an example

PR #234, opened 2019-07-24
Fix bootstrap header

### LINKED ISSUE #227: Running bootstrapped script fails on bare Ubuntu
1. Create script with bootstrap header using: `kscript --add-bootstrap-header some_script.kts` 

2. Start bash in bare ubuntu 18.04 using docker: `docker run -it -v "$(pwd):/vol" ubuntu:18.04 bash`
3. Run your bootstrapped script
```bash
cd /vol
./some_script.kts
```

It will fail on several steps:
1. `./some_script.kts: line 5: curl: command not found`
2. (after installing curl): 
    ```
    Looking for unzip...
    Not found.
    ```
3. (after installing unzip)
    ```
    Looking for zip...
    Not found.
    ```
4. (after installing zip)
    ```
    $ source $SDKMAN_DIR/bin/sdkman-init.sh
    main: line 3: /bin/sdkman-init.sh: No such file or directory
    ```
5. (after: `source "/root/.sdkman/bin/sdkman-init.sh"`)
    ```
    ./some_script.kts: line 6: exec: kscript: not found
    ```
6. (after: `source "/root/.sdkman/bin/sdkman-init.sh"` again)
    ```
    /root/.sdkman/candidates/kotlin/current/bin/kotlinc: line 80: java: command not found
    ```
    
In docs we can read: 
> Now some_script.kts can be shared and run directly on any other machine **that has bash**, without having to go through the Installation steps first.

From my experience described above, it is simply not true. 

Implicit dependency on basic tools like `curl` or `zip` is not a big deal,
but requiring `java` on the host clearly contradict statement from the docs. 

Perfect solution would be: make bootstrapped script work on any Ubuntu host which has `bash` installed (and `curl`, `zip`, `unzip` also - because I suppose it would be very hard to not depend on them, especially on `curl`) 
Good enough solution would be: list all required dependencies in the docs.

Regardless of that, the example above reveals also other problem: depending on variable `SDKMAN_DIR` which is not set.

0legg:
Fixes #227 resolving following issues:
- Misprint in quotes mark lads to fail after installing sdkman — command resolves to `$SDKMAN_DIR/bin/sdkman-init.sh` instead of `/home/olegg/.sdkman/bin/sdkman-init.sh`
- Missing java installation routine — when running `kscript` on a fresh Ubuntu installation, script crashes with error message `/home/olegg/.sdkman/candidates/kotlin/current/bin/kotlinc: line 80: java: command not found`
- Failure on script re-run — if `sdkman` already had `kscript` installed, it was erring out instead of silently continue.
- Failure on initial install — after all required tools are installed, we should re-source sdkman, so shell will see all newly installed tools.

### DIFF
--- a/misc/boostrap_header.sh
+++ b/misc/boostrap_header.sh
@@ -3,12 +3,14 @@ _in_path kscript && echo "kscript is already installed at $(which kscript)" 1>&2
     function echo_and_eval() { echo "$ $@" 1>&2; eval "$@" 1>&2; }
     _in_path sdk || {
         curl "https://get.sdkman.io" | bash 1>&2 && \
-            echo_and_eval source "$"SDKMAN_DIR/bin/sdkman-init.sh
+            echo_and_eval source "$SDKMAN_DIR/bin/sdkman-init.sh"
     }
 
     sdkman_auto_answer=true
 
+    _in_path java || echo_and_eval sdk install java
     _in_path kotlin || echo_and_eval sdk install kotlin
     _in_path gradle || echo_and_eval sdk install gradle
-    echo_and_eval sdk install kscript
+    _in_path kscript || echo_and_eval sdk install kscript
+    echo_and_eval source "$SDKMAN_DIR/bin/sdkman-init.sh"
 }


0legg:
Perfectly, I'd rename the file to fix misprint (now it's called `BOOstrap`, but it's nothing scary in it), but it'll require short link update.

0legg:
Regarding Travis CI: failing test is red on master as well (ref #219)

holgerbrandl:
Thanks for the nice PR. 

I'm still a bit unsure if installing java silently @on the user machine is a sound practice, but at least it makes scripts much more self-contained.

I'll try to address the failing tests on master asap.

PR #248, opened 2019-11-12
Add "Authentication from environment"

alekseytomin:
I have to use the private nexus repository in gitlab-ci.
We use only authenticated access to it. But I mustn't write a password in scripts.

PS: I've chosen '@' as a symbol ('$' isn't possible in Kotlin). Maybe You will choose another one.

PPS: I have not found any other point to use environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,10 +316,10 @@ Using annotations instead of comment directives to configure scripts is cleaner 
 
 // also protected artifact repositories are supported, see <https://github.com/holgerbrandl/kscript/blob/master/test/TestsReadme.md#manual-testing>
 // @file:MavenRepository("my-art", "http://localhost:8081/artifactory/authenticated_repo", user="auth_user", password="password")
-// You can use environment variables for user and password
-// @file:MavenRepository("my-art", "http://localhost:8081/artifactory/authenticated_repo", user="@artifactory_user", password="@artifactory_password")
+// You can use environment variables for user and password when string surrounded by double {} brackets 
+// @file:MavenRepository("my-art", "http://localhost:8081/artifactory/authenticated_repo", user="{{artifactory_user}}", password="{{artifactory_password}}")
 // will be use 'artifactory_user' and 'artifactory_password' environment variables
-
+// if the value doesn't found in the script environment  will fail
 
 // Include helper scripts without deployment or prior compilation
 @file:Include("util.kt")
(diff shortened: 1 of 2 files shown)


holgerbrandl:
Great suggestion. I would also never include credentials in scripts.

Indeed @ feels a bit odd. Maybe `{{MY_PW}}` could feel a bit more _templaty_?

I think before we should merge it, the fallback behaviour when a variable is not defined should be changed to make kscript stop script execution. Tis may clash with the rare cases that a pw is actually starting with {{ and ends with }}, but it seems preferable to be than using wrong credentials, just because the user forgot to `export` the variables in bash before invoking kscript.. What do you think?

alekseytomin:
> Great suggestion. I would also never include credentials in scripts.
> 
> Indeed @ feels a bit odd. Maybe `{{MY_PW}}` could feel a bit more _templaty_?

Ok. I've forgotten this.

> I think before we should merge it, the fallback behaviour when a variable is not defined should be changed to make kscript stop script execution.

`System.getenv()` in Kotlin is a nonnull function.
Without any changes we have an exception:
```Exception in thread "main" java.lang.IllegalStateException: System.getenv(value.subs…ing(2, value.length - 2)) must not be null
	at kscript.app.DependencyUtilKt.decodeEnv(DependencyUtil.kt:72)
...
```
I think better to write something like
```Exception in thread "main" java.lang.RuntimeException: environment variable 'HEXUS_USER' doesn't found
	at kscript.app.DependencyUtilKt.decodeEnv(DependencyUtil.kt:76)
...
```

Or maybe hide the stacktrace?

holgerbrandl:
3 minor aspects (which I typically would just do myself, but I'm a bit in a hurry this week)
1) In the docs the example pw artifactory_password should be uppercase ARTIFACTORY_PASSWORD since bash variable are typically uppercase
2) I like the more custom error. But the wording seems slightly wrong. Couldn't we just say

> "Could not resolve environment variable {{bla}} in maven repository credentials"

3) Instead of rethrowing the exception, we should better exit with 
```
errorMsg(lazyMessage().toString())
quit(1)
```
This will prefix the message with `[kscript] [ERROR]` which will help the user to distinguish script logic errors from kscript runtime errors.

alekseytomin:
Done

alekseytomin:
https://github.com/holgerbrandl/kscript/pull/248/commits/c8fc911e05009936e1697cc188a8794ff9dd4292
I've found than `System.getenv(key)` has a wrong type (`String!` instead of `String?`) and remove from code. `System.getenv()[key]` returns `String?`

alekseytomin:
Is it okay?

holgerbrandl:
Great PR, thanks for your support and patience. A release including your feature is planned for later today.

alekseytomin:
Thanks!

PR #252, opened 2019-12-30
Add missing inputs to the incremental task "test"

theosotr:
Hi

This pull request fixes the incremental task `test`.

This task depends on some test resources included in the `test/resources` directory.
However, this directory is not declared as input to this task, so tests are not executed whenever there are updates to any of those dependent files.

This commit fixes this issue by declaring `test/resources` as input of task `test`.


holgerbrandl:
Thanks for your contribution.

Issue #222, opened 2019-04-10
Consider switching to Kotlin DSL for generated `--idea` flag


mikeholler:
Command:

```
kscript --idea some-file.kts
```

Produces an IntelliJ project with a `build.gradle` in the root. Since we all love Kotlin, it may make sense to transition this to a `build.gradle.kts` now that's got first party support in Gradle.

Thanks for such a great project!

holgerbrandl:
Sounds like a great idea for a PR to me. Would you be willing to contribute it?

holgerbrandl:
Any news, or shall we close this issue?

mikeholler:
Hey @holgerbrandl, I don't personally have the time to contribute this change, and only wanted to suggest it here in case someone else wanted to take it up. In projects I manage, I generally leave tickets open if we still think it's a good idea so that others can come along and find something they may want to contribute. If you'd like to close it instead though, that's completely up to you.

Thanks for responding :smile:

martinbonnin:
Fixed with https://github.com/holgerbrandl/kscript/pull/258

PR #257, opened 2020-02-22
Generate a build.gradle.kts file

martinbonnin:
Switch the gradle file generated by the `--idea`  option to use Kotlin.

Closes https://github.com/holgerbrandl/kscript/issues/222


src/main/kotlin/kscript/app/AppHelpers.kt
```diff
@@ -255,16 +262,16 @@ $stringifiedRepos
 }
```
martinbonnin: Not directly related to this PR but I noticed the `build.gradle` file includes `mavenLocal()` by default. Is that intended ?

martinbonnin:
Merged at the same time as https://github.com/holgerbrandl/kscript/pull/258

Issue #188, opened 2018-10-31
Using @file:CompilerOpts("-jvm-target 1.8") doesn't work for --idea


jsonbrooks:
running kscript --idea some_script.kts doesn't put the option in the gradle build file for the resulting temporary project.

holgerbrandl:
It's unlikely that we can implement a mirroring of all possible compiler options onto gradle settings. Feel provide to provide a PR at least the most common one, which is indeed the jvm target.

PR #258, opened 2020-02-22
Forward jvm target to --idea generated build.gradle.kts

martinbonnin:
This pull request is based on https://github.com/holgerbrandl/kscript/pull/257 so it includes more commits that strictly needed.

It allows to have IDE support for Java 8 and more if declared in the Compiler options.

Closes https://github.com/holgerbrandl/kscript/issues/188


src/main/kotlin/kscript/app/AppHelpers.kt
```diff
@@ -275,15 +318,15 @@ sourceSets.test.java.srcDirs 'src'
 
         // also symlink all includes
         includeURLs.distinctBy { it.fileName() }
-          .forEach {
+            .forEach {
```
martinbonnin: This came after auto-reformating code. I have my IDE set on tab size = indent = continuation = 4 spaces. Should I use something else ?

holgerbrandl:
Thanks @martinbonnin for this great contribution!

Issue #261, opened 2020-03-18
Maven repo credentials are no injected in gradle.build when using "--idea" flag


jiraguha:
Hi, Thanks for this awesome tool :-). I have a little bug:
When in *.kts you have
```
@file:MavenRepository("id-repo","url-repo", password = "pass", user="user")
```
And use the flag `--idea`, the generated gradle.build.kts is:

```
repositories {
    ...
    maven {
        url = uri("url-repo")
    }
}
```
But it should be

```
repositories {
    ...
    maven {
        url = uri("url-repo")
        credentials {
            username = "user"
            password = "pass"
        }
    }
}
```
Kr,
JP

holgerbrandl:
Indeed a bug. Thanks for spotting it.

holgerbrandl:
Fixed with #262

PR #262, opened 2020-03-18
Add maven repo credentials in build.gradle.kts generated by --idea

jiraguha:
Possible solution for the issue #261


holgerbrandl:
Thanks for your contribution. Simple but effective fix (which also fixes #261)

PR #264, opened 2020-04-08
Add option to launch temp intellij as custom command through env property

matheus-teles:
## Why
Depending on how your intellij was installed it has not the same command set to launch, so this PR allows to tell kscript what command needs to be considered

## What
Allow the possibility to customize kscript to set your intellij launch command


holgerbrandl:
Thank you very much for your contribution to kscript!

PR #265, opened 2020-05-26
bump kotlin version to 1.3.72

martinbonnin:
No specific reason other than not having to download too many kotlin versions.


holgerbrandl:
Thanks for the help.

PR #268, opened 2020-07-08
Fix idea project symlinks for complex projects

rossbacher:
Create symlinks for included files in the same relative path to the main script that they are in the actual script location.

e.g. right now if you have 

**`Main.kt`**:

```kotlin
#!/usr/bin/env kscript
@file:Include("./directory/Include.kt")

fun main(args: Array<String>) {
    println(helloWorld)
}
}
```

**`directory/Include.kt`**:

```kotlin
#!/usr/bin/env kscript

val helloWorld = "Hello World!"
```

in the idea project they will end up like this:

```
├── build.gradle
├── gradle
│   └── wrapper
│       ├── gradle-wrapper.jar
│       └── gradle-wrapper.properties
├── gradlew
├── gradlew.bat
├── local.properties
└── src
    ├── Include.kt -> /Users/arossbacher/Documents/directory/Include.kt
    └── Main.kt -> /Users/arossbacher/Documents/Main.kt
```

The idea project itself will not be able to run the code as the referenced file is at the wrong positions relative to the main file.

with this fix it looks like this:

```
├── build.gradle.kts
├── gradle
│   └── wrapper
│       ├── gradle-wrapper.jar
│       └── gradle-wrapper.properties
├── gradlew
├── gradlew.bat
├── local.properties
└── src
    ├── Main.kt -> /Users/arossbacher/Documents/Main.kt
    └── directory
        └── Include.kt -> /Users/arossbacher/Documents/directory/Include.kt
```

the directory structure is kept the same and thus this is working in the IDE.

This allows the script in the idea workspace to be fully operational even when including relative path includes and e.g. enable debugging.

### DIFF
--- a/src/main/kotlin/kscript/app/AppHelpers.kt
+++ b/src/main/kotlin/kscript/app/AppHelpers.kt
@@ -333,7 +333,7 @@ $kotlinOptions
         // also symlink all includes
         includeURLs.distinctBy { it.fileName() }
                 .forEach {
-                    val includeFile = when {
+                    val symlinkSrcDirAndDestination = when {
                         it.protocol == "file" -> {
                             val includeFile = File(it.toURI())
                             val includeDir = Paths.get(includeFile.path).parent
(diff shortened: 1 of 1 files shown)


holgerbrandl:
Nice and reasonable improvement. Thanks for contributing it.

PR #269, opened 2020-07-13
Use hard links for idea project 

rossbacher:
### Use hard links for idea project as Kotlin compiler has issues with symlinks.

The second run of `./gradlew assemble` in generated idea project after a code change always let to

```
> Task :compileKotlin
e: <.kt file path in idea project>: Redeclaration: ClassA
e: <.kt file path in original location>: Redeclaration: ClassA
```
This can only be fixed by cleaning the project or deleting the `build` folder. Which is really annoying when working on a project.

This can be avoided using hard links, plus I see no downside of using hard links here as they would only come to affect if it would be common to copy/replace a different file with the same name over the original project files as a regular use case, which it isn't.

### DIFF
--- a/src/main/kotlin/kscript/app/AppHelpers.kt
+++ b/src/main/kotlin/kscript/app/AppHelpers.kt
@@ -327,7 +327,7 @@ $kotlinOptions
         mkdir()
 
         // https://stackoverflow.com/questions/17926459/creating-a-symbolic-link-with-java
-        createSymLink(File(this, scriptFile.name), scriptFile)
+        createLink(File(this, scriptFile.name), scriptFile)
         val scriptDir = Paths.get(scriptFile.path).parent
 
         // also symlink all includes
@@ -346,7 +346,7 @@ $kotlinOptions
                             Pair(this, fetchFromURL(it.toString()))
                         }
                     }
-                    createSymLink(File(symlinkSrcDirAndDestination.first, it.fileName()), symlinkSrcDirAndDestination.second)
+                    createLink(File(symlinkSrcDirAndDestination.first, it.fileName()), symlinkSrcDirAndDestination.second)
                 }
     }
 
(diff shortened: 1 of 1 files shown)


holgerbrandl:
What about typical linux servers where there are different disks? E.g. one for data and one for home-directories. Wouldn't hardlinks fail if a script is located on a secondary disk with a separate filesystem compared to the homedirectory FS (where kscript would setup the temporary project)?

Also see (more as reminder for me I guess https://blog.usejournal.com/what-is-the-difference-between-a-hard-link-and-a-symbolic-link-8c0493041b62)

rossbacher:
I did not think about the typical spread in Unix systems, we use Macs here and ci is Linux but all is on one partition. Those are good points. 

I still have the issue that a second build in IJ has all these duplicate files if symlinked. I will investigate a bit more to understand what exactly is causing it and if there is a different way of working around it.

Also would you be open to a flag for which links to use? As if I don't find a general workaround this might still be useful for some users. (I've been working with a version built of this branch for some time (on our scripts).

holgerbrandl:
Multi disk rack servers are common at least in the few orgs I was working in. So I guess we should support this usecase as well. 

I still struggle to reproduce the problem under ubuntu:

```
echo 'println("foo")' > test.kts
kscript --idea test.kts
```
Next, I pick up the temp path from the ide and run gradle build in the terminal:

image
This works. Can you help me to reproduce the problem?

Depending on the severity, the hardlink mode could/should be enabled with flag or become the new default. The latter should just apply if the majority of users is affected by the problem.

rossbacher:
Oddly enough it is gone for me now too and I have no idea how this was happening for me. 🤷‍♂️ 

What i was doing is use `--idea` to create the Idea project and then on the `main` function there is the green play button in the gutter and i just clicked it (gradle import worked automatically). And that then created a run configuration in IJ (actually I think I used Android Studio for that and was just testing). 
And with that configuration I could only use it until I made a code change. After the code change I had those odd duplications between the location of the link and the destination of the link (as if it was compiling both locations). I needed to clean the project to get it compiling again.

Somehow now I cannot create a run configuration automatically from the gutter anymore and if I create a `Kotlin` run configuration manually for that script (and select the class with the main function manually) it just works including me changing code and recompiling. 🤷‍♂️ 

I'll close this PR for now as I also found other odd behaviors when using hard links (e.g. the actual file in our project not always getting updated).

Issue #270, opened 2020-07-27
[Bootstrap Script] The SDKMAN_DIR is not set / not inherit from SDKMAN installation script


patrickyin:
I've got this error message when I run a script which contains kscript's bootstrap script:
```bash
# ...
$ source /bin/sdkman-init.sh
bash: line 3: /bin/sdkman-init.sh: No such file or directory
$ sdk install kotlin
bash: line 3: sdk: command not found
$ sdk install gradle
bash: line 3: sdk: command not found
$ sdk install kscript
bash: line 3: sdk: command not found
$ source /bin/sdkman-init.sh
bash: line 3: /bin/sdkman-init.sh: No such file or directory
```

These messages were coming up after the SDKMAN installed.

holgerbrandl:
Fixed by PR

PR #271, opened 2020-07-27
Fix bootstrap script env variable issue

patrickyin:
### Issue
Error messages in: #270

In `misc/boostrap_header.sh`, it installs SDKMAN using
```bash
curl "https://get.sdkman.io" | bash 1>&2 && \
            echo_and_eval source "$SDKMAN_DIR/bin/sdkman-init.sh"
```
After that, it installs Java, Kotlin, Grade, and Kscript. 
It uses `$SDKMAN_DIR`, but it can't be retrieved from the previous sdkman installation script. so it causes the issue: `misc/boostrap_header.sh` runs `/bin/sdkman-init.sh` since `$SDKMAN_DIR` is not set. In the meantime, the rest of the installation will be failed.

### Fix
I added the declaration of the `SDKMAN_DIR` before SDKMAN installation script runs. The value is same as the one SDKMAN script uses.
There are 2 reasons I declare it explicitly:
* The SDKMAN installation script will use it as the SDKMAN's location. So whatever the default location is changed or not in SDKMAN side, we will always install SDKMAN to `$HOME/.sdkman`
* The rest of the installation will be able to consume this `SDKMAN_DIR` immediately.

### DIFF
--- a/misc/boostrap_header.sh
+++ b/misc/boostrap_header.sh
@@ -2,7 +2,7 @@ function _in_path() { command -v "$1" >/dev/null 2>&1; }
 _in_path kscript && echo "kscript is already installed at $(which kscript)" 1>&2 || {
     function echo_and_eval() { echo "$ $@" 1>&2; eval "$@" 1>&2; }
     _in_path sdk || {
-        curl "https://get.sdkman.io" | bash 1>&2 && \
+        export SDKMAN_DIR="$HOME/.sdkman" && curl "https://get.sdkman.io" | bash 1>&2 && \
             echo_and_eval source "$SDKMAN_DIR/bin/sdkman-init.sh"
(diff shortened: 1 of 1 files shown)


patrickyin:
cc: @holgerbrandl

holgerbrandl:
Thank @patrickyin for the PR. Please give me a little more time to review and merge it.

patrickyin:
Hi @holgerbrandl , have you had a chance to work on this PR? Waiting for merge to make my friend's local script work. : ) Cheers.

holgerbrandl:
Thank you for the PR. I still feel unsure about the self-bootstrap feature since it has zero test coverage, but I have to blame myself here for being so sloppy in the first place.

PR #275, opened 2020-08-13
Update README.md

milosgarunovic:
Got error on the changes line when running `kscript --idea CountRecords.kts`
`[kscript] [ERROR] Could not read script argument 'CountRecords.kts'`

Without extension, it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,7 +417,7 @@ Boostrap IDEA from a `kscript`let
 Artifacts and versions will differ between scripts, so it is hard to maintain them all in a single project. To nevertheless provide optimal tooling when scripting with Kotlin `kscript` allows to create temporary projects for `<script>` arguments. .
 
 ```bash
-kscript --idea CountRecords.kts
+kscript --idea CountRecords
 ```
 This will open [IntelliJ IDEA](https://www.jetbrains.com/idea/) with a minimalistic project containing just your (1) `<script>` and (2) a generated `gradle.build` file:
 


holgerbrandl:
If the script is named CountRecords.kts the documented syntax is correct.

milosgarunovic:
So `kscript --idea CountRecords.kts` opens the existing script in IntelliJ, but `kscript --idea CountRecords` will create a new script and open it in IntelliJ. I just tried that and got:

```
> kscript --idea CountRecords
[kscript] Setting up idea project from /var/folders/p9/jv3_6j6929vd3d0q0dr8vyd80000gn/T/tmp2649147066000859856.tmp/scriptlet.e93f994f01c537c4.kts
[kscript] Project set up at /Users/milosgarunovic/.kscript/kscript_tmp_project__scriptlet.e93f994f01c537c4.kts_1598246065629
> kscript --idea CountRecords.kts
[kscript] Setting up idea project from CountRecords.kts
[kscript] Project set up at /Users/milosgarunovic/.kscript/kscript_tmp_project__CountRecords.kts_1598246100454
```

The first call is for creating a script, second is to open from existing .kts file.
This is to be expected than, maybe to update docs so people know the difference?

holgerbrandl:
It should not create a missing script but throw an error instead. Sorry for the confusion.

Let's keep this open until I manage to implement a fix.

Thanks for spotting it.

milosgarunovic:
No problem, just thought I could contribute to this, at least to help people set up new scripts (which I was after). Thank you for the comments 😄

holgerbrandl:
It correctly throws an error if the non-existing script file ends with kt or kts
```
kscript --idea huhu.kts
[kscript] [ERROR] Could not read script argument 'huhu.kts'
```

Since kscript also supports direct script arguments (e.g. `kscript 'println("test")'`) it is hard to infer if a given string corresponds to an intended file name or to a direct script argument (see https://github.com/holgerbrandl/kscript#script-input-modes). E.g. consider `kscript Integer.MAX_VALUE` which is a legit program evaluating the MAX_VALUE internally without printing it. To disambiguate this a simple heuristic such as 
```
fun isFileNameCandidateWithoutKSuffix(file: File): Boolean {
    return file.name.matches("[A-z][A-z_.-]*".toRegex())
}
``` 
is not sufficient because we'd need to consider legit single word language constructs here.

Thus, I'm inclined to keep the behavior as it is for now. Sorry for any inconvenience this might cause you.

milosgarunovic:
Makes sense, nice explanation :) No inconvenience, I just thought it works differently, but it makes sense now that you say that it can execute code instead of file. Thank you for your time, I really appreciate it!

PR #284, opened 2020-10-21
Add Kotlin run configuration for .kt based scripts plus add gradle wrapper.

rossbacher:
When you run with the `--idea` option `kscript` is creating an IDEA run configuration for your script.

If you have a `.kts` script it makes sense to create a `bash` type run configuration. But those do not work if your script is a `.kt` script with a `main` method.

This change will create a Kotlin run configuration in that case.

This will also add a gradle wrapper to the tmp project used for editing the script in IDEA.


holgerbrandl:
Wonderful addition. Thank you so much for your continued support. :-)

PR #287, opened 2020-10-24
Update README.md

FreekDB:
Small modification of sdkman install commands:
source "$HOME/.sdkman/bin/sdkman-init.sh"

(see https://sdkman.io/install)


holgerbrandl:
Thank you for this nice correction.

PR #295, opened 2020-12-08
Fix Java info in `kscript --version`

erikhuizinga:
When `kotlin -java` contains Java version output like so:

```
Kotlin version 1.4.20-release-308 (JRE 15.0.1+9)
```

then `kscript --version` shows

```
Java      : JRE 15.0.1+9)
```

Note the last character, a closing parenthesis that shouldn't be there.
This commit fixes that issue.


holgerbrandl:
I always love and value this sense for tiny details. Thanks for the fix.

PR #297, opened 2020-12-14
Add error message for MavenRepository

sunyal:
- improve the error message to include a message in the stacktrace indicating what is the problem


holgerbrandl:
Thank you @sunyal for this nice addition to provide a more seamless user experience.

PR #315, opened 2021-04-23
update user guide testing on ci to include a github action section

plusmobileapps:
# Overview 

Document how to use a GitHub Action for running unit tests in pull requests against your main branch 

## Description 

The [helper script](https://gist.github.com/elihart/019bd116d3fa0d6214b7396eedc4b206) by [@elihart](https://gist.github.com/elihart) was really helpful for running unit tests locally. But I was wanting to create an even easier way to run this on CI using GitHub Actions so I created a new action called [kscript-action](https://github.com/plusmobileapps/kscript-action) which runs that helper script in a docker container. I was able to use this in a [sample project](https://github.com/plusmobileapps/kotlin-scripting) and validated that it worked. 

There was some flakiness in the script sometimes timing out so maybe in a future iteration can allow the user to pass a timeout argument in the action? But usually rerunning it once got the script to run fine. Open to any feedback or contributions on that GitHub Action as its the first one I have ever written.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -163,6 +163,30 @@ Note that the Gradlew wrapper is not automatically created by `kscript --idea` s
 
 To streamline testing there is also [helper script](https://gist.github.com/elihart/019bd116d3fa0d6214b7396eedc4b206) by [@elihart](https://gist.github.com/elihart) that runs `gradle clean test` on other scripts to easily build and test them (using kscript). This script takes the name of another kts script as an argument, builds that script with gradle, and runs its tests via gradle clean test.
 
+#### Github Action 
+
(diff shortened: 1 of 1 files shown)


holgerbrandl:
Great idea. I'm using github actions, but I'm not that familar with the technical concepts/workflow underpinning them. When glancing through https://github.com/plusmobileapps/kscript-action/blob/main/Dockerfile I wondered if it should rather use a specific version to build the image? As of now, the image being built would change when a new version of kscript is being released. Did I miss something here?

plusmobileapps:
@holgerbrandl  that is a valid point, I did notice in the [kscript Dockerfile](https://github.com/holgerbrandl/kscript/blob/master/misc/Dockerfile#L26) that there is a kscript version that was declared, although it was a little mysterious to me where that version was ever being supplied. I think it is a good idea though to allow the users to be able to specify the kscript version for consistency. It might help me if you [create an issue](https://github.com/plusmobileapps/kscript-action/issues/new) on the action for what you were looking for exactly and I can try to implement it. 

This was also the first Dockerfile that I have ever written so its been a learning experience 😄

holgerbrandl:
This mystery I can help to solve easily, the version is supplied in the release workflow in https://github.com/holgerbrandl/kscript/blob/master/misc/kscript_release.sh#L291
when building the image

Regarding the action, I guess supplying a kscript version parameter with the current as default would imho be the most sensible solution.

PR #318, opened 2021-05-26
Kscript.kt: Fix whitespace in usage message

jayvdb:


holgerbrandl:
Thanks for the correction.

PR #325, opened 2021-08-11
Use alpine-based image for Dockerfile:

sfesenko:
- reduce image size 217 Mb vs 544Mb
 - able to referer local files via volume mountig:
    `docker run -v $PWD:/w  kscript ./clikt-example.kts --count 2 --name Hey`


holgerbrandl:
Great PR. Not just the base image was updated, but you also managed to refactor out sdkman entirely. Great work, thanks for sharing/contributing!

Issue #324, opened 2021-08-11
Bootstrap headers not working (bare centeos)


robert2411:
When I run a simple script with the bootstrap headers I get the folloing error:
`./test.kts: line 6: exec: kscript: not found`

I created a reproducable in docker

```
FROM centos
RUN yum install unzip -y
RUN yum install zip -y
WORKDIR /project
COPY ./test.kts /project
RUN chmod +x test.kts
RUN ./test.kts
```

the script
```
#!/bin/bash

//usr/bin/env echo '
/**** BOOTSTRAP kscript ****\'>/dev/null
command -v kscript >/dev/null 2>&1 || curl -L "https://git.io/fpF1K" | bash 1>&2
exec kscript $0 "$@"
\*** IMPORTANT: Any code including imports and annotations must come after this line ***/


println("Hello from Kotlin!")
for (arg in args) {
    println("arg: $arg")
}
```

holgerbrandl:
Hmm,  I can't reproduce the problem yet. I've uninstalled kscript and ran your script, which works fine: It installs kscript via sdkman and runs the script afterwards.

The install procedure is pulled from https://raw.githubusercontent.com/holgerbrandl/kscript/master/misc/boostrap_header.sh so maybe this fails in your environment?

holgerbrandl:
It seems that the bootstrap header is not correct. 

Currently with
```
#!/bin/bash

//usr/bin/env echo '
/**** BOOTSTRAP kscript ****\'>/dev/null
command -v kscript >/dev/null 2>&1 || curl -L "https://git.io/fpF1K" | bash 1>&2
exec kscript $0 "$@"
\*** IMPORTANT: Any code including imports and annotations must come after this line ***/
```
it can't find kscript because it has not sourced the init-script from SDK. I think `source "$SDKMAN_DIR/bin/sdkman-init.sh"` is missing in before exec. What do you think?

wuseal:
I saw the log that I pasted, It has sourced the sdk script:
```
Setting kscript 3.1.0 as default.
$ source /Users/user/.sdkman/bin/sdkman-init.sh
./test.kts: line 6: exec: kscript: not found
```
But also error with can't find `kscript`

wuseal:
I fixed it by change the bootstrap header to next:
```
//usr/bin/env echo '
/**** BOOTSTRAP kscript ****\'>/dev/null
[...]

PR #330, opened 2021-10-01
Including scripts by relative reference - Fixes #303

### LINKED ISSUE #303: Including scripts by relative reference
I have seen issue #218 and from that ticket, it looks like the problem is already solved, but in my opinion, it's not, or at least not in all cases.

Please consider the following directory structure:
obraz

**service-script-a.kts** includes **common-script-a.kts**. But **common-script-a.kts** includes internally **common-script-b.kts**.
Now when I execute **service-script-a.kts** I get error: 

`[kscript] [ERROR] Failed to resolve //INCLUDE 'common-script-b.kts'
[[kscript] [ERROR] /home/vagrant/workspace/Kod/GitRepos/TestCase/service-a/common-script-b.kts (No such file or directory)]`

Fixing it shouldn't be difficult - it is enough to include files **always** relatively to the file which does the include. Without that, it is not possible to correctly structure scripts and encapsulate pieces of code.

I am attaching the above very simple test-case as a zip file.

[TestCase.zip](https://github.com/holgerbrandl/kscript/files/5921348/TestCase.zip)

aartiPl:
This patch Fixes #303 

I have tested it on my branch with URLs in the includes directory directed to my branch. After that, I have reverted them back to point to your master branch. Because of that, tests will not work on the branch, but only on the master.

I have started only Tests class in src/test directory. I am not sure how to start other tests. Please let me know if something else should be done to merge this PR.


holgerbrandl:
Awesome. Thank you so much for supporting script!

I've run the test but they still seems to struggle with the PR:

Tests > test_include_annotations FAILED
    org.junit.ComparisonFailure: expected:<...ion
    fun url_included[_1() = println("i came from the internet")
    fun url_included_2() = println("i also] came from the inter...> but was:<...ion
    fun url_included[() = println("i] came from the inter...>
        at io.kotlintest.matchers.MatchersKt.shouldBe(Matchers.kt:15)
        at Tests.test_include_annotations(Tests.kt:212)

Tests > test_include_detection FAILED
    java.lang.AssertionError: expected: 2 but was: 1
        at io.kotlintest.matchers.MatchersKt.equalsError(Matchers.kt:89)
        at io.kotlintest.matchers.MatchersKt.shouldEqual(Matchers.kt:63)
        at io.kotlintest.matchers.MatchersKt.shouldBe(Matchers.kt:55)
        at Tests.test_include_detection(Tests.kt:220)

aartiPl:
There is an inherent problem with testing changes covering resolving URLs on branches, at least in the current test setup. Test files rely on KScript Github *master* branch repository. But the problem is that if you make changes in test files on a branch, then those files are not yet in the *master* branch of the Github repo, so the tests will just fail.

In case of this PR, there is added a new script file, which is included by URL in test/resources/includes/rel_includes/include_by_url.kt. This new file is obviously not in the master branch, because PR is not yet merged, and that's why tests on the branch fail.

In such a case, to accomplish the tests on the PR branch you can do the following things:

1) Modify URLs to the test branch in files:
test/resources/includes/include_by_url.kt
test/resources/includes/include_variations.kts

For testing in my forked branch I have used the following URLs:
include_by_url.kt: @file:Include("https://raw.githubusercontent.com/aartiPl/kscript/Including_scripts_by_relative_reference_-_fixes_%23303/test/resources/includes/rel_includes/include_by_url.kt")
include_variations.kts: @file:Include("https://raw.githubusercontent.com/aartiPl/kscript/Including_scripts_by_relative_reference_-_fixes_%23303/test/resources/includes/include_by_url.kt")

2. Otherwise the tests can be run successfully only on the "master" branch, so only after the merge of PR.

holgerbrandl:
Packaging seems indeed broken currently (see #324). 

I've run the tests locally against the PR and struggle to complete "kscript https://git.io/fxHBv" both on the terminal and from the IDE. This is because the file-suffix is no longer detected correctly. It must be kt but seems to be kts with PR. Am I missing something here?
Make sure to always wipe the cache before testing.

aartiPl:
There is indeed something wrong with my implementation. I will try to figure out what's that and provide a fix. Thanks for your input!

aartiPl:
Ok. I have managed to fix the problem. The remaining problems are with:

* packages - seems that they are broken anyway
* idea project - I can not test it as I am working in a virtual machine without UI

obraz
obraz

Please check if idea generated project is working as expected.

It seems to me that there would be needed some deeper refactoring in KScript. I have an idea how to do that. If you are willing to do such changes, I can work on it. But I prefer to do this after applying changes from this PR first.

aartiPl:
Seems that the last broken use case is fixed now. There are failures only with packaging and idea. Can you please check those use cases?

holgerbrandl:
Thank you so much @aartiPl  for your patience with my slow response times, and many thanks for your great support.

PR #331, opened 2021-10-11
Support for running bash commands on Windows (via sh instead of bash)

pmatysek:
Kotlin was in my PATH, bu I was getting error `[kscript] [ERROR] kotlinc is not in PATH`. So I've made some investigation and testing. As I figured out problem was with `bash -c` command. For some reason it's not showing any output neither while running on my windows cmd nor from kscript. There wasn't any output even from simple `echo`. Fortunately I found that command `sh -c` is working perfectly.
For that reason I wrote some piece of code which is choosing proper bash command for OS.
However I'm not sure if problem is only in my enviroment (with git-bash instalation or some windows configuration?) or global for windows.


aartiPl:
The solution is now applied in similar form on kscript_4.1 branch. Closing this PR because of that.

PR #332, opened 2021-11-21
Redesign of kscript

aartiPl:
Hi!

This is a redesign of kscript. The main goal is to split kscript into smaller pieces based on object-oriented design to allow easier modifications of kscript in the future. 

This rewrite is not yet finished (since draft PR), but most of the cases are already working correctly. I will prepare a detailed changelog later, but for now, it would be nice if you can review the changes.

Currently, only the new UT do not work, but the whole test_suite is working as previously (packaging and idea tests are failing).

This is a *huge* refactor, but please have a look - I believe it is worthy to apply it as it will make modifications much, much easier.
Not yet finished as I mentioned before, but if you like it I am able to devote some more time to accomplish changes. 

Probably a candidate for version 4.0-beta because of the scope of changes.


holgerbrandl:
Thanks @aartiPl for this hard and quite extensive work. I'll will look into it, but usually, I only find time during the weekends to do so. So please be a bit patient with me.

Apart from being more modular, what functional benefits (new features) does the PR provide?

aartiPl:
Yes - I have the same. Usually, I can not spend so much time on my projects :-|

I have put a summary of the changes into CHANGES.md file. It is mostly refactoring, but changes in the structure of code are important to any future changes of app and improving code coverage.

Please apply "Squash Merge" - there is so many experiments in my changes history, that it doesn't make sense to keep it in Git

holgerbrandl:
It's a wonderful rewrite of the entire application. Thanks a ton. 

Do you see any issues which we should try solving on the PR-branch before merging it back into master? Do you have any suspicion regarding the few still failing tests?

Clearly, we should level up the version to 4.0.

Would you mind/would it help if I add you as a committer to the repo?

aartiPl:
Great that you like the rewrite! :-)

There are a few points which should be addressed sooner or later:

1) One of the idea test cases (include_variations.kts) is failing because there are two files that have the same name, although they come from different places in the file system. Because of that during the creation of the project, there is an exception thrown:  java.nio.file.FileAlreadyExistsException: /home/vagrant/.kscript/cache/idea_b8ddf3b5ac635b9dfcbfce05e39e9635/src/include_by_url.kt. The simplest way to solve that is just to rename the file with some random suffix, whenever there is an already existing file with the same name in the project. I am pretty sure that this test case was not working before anyway - I have added the file with the same name, but a different place in the directory structure to the test suite previously, to test recursive inclusion of files.
2) Gradle build files created by the app should have proper indentation - currently, they are correct, but indentation is a bit screwed up.
3) Packaging is not working, but most probably it should be possible to revive it by using the current method of packaging.
4) I am thinking about adding caching of classpath again. I can do some additional tests to make sure that there is a noticeable difference in performance and add caching of classpath in a similar way as it was before. 
5) I couldn't update kotlin-scripting-jvm to v1.5.31 because in that version there is a bug, which prevents the correct resolution of dependencies if they do not exist in the local repository. So now most of the app is compiled with 1.5.31 (needed e.g. for value classes or list builders), but the resolution of dependencies is done through old kotlin-scripting-jvm v1.4.32. So this is not a perfect solution, but seems to work well enough. Alternatively, instead of kotlin-scripting implementation it is possible to use Maven Archeologist: https://github.com/square/maven-archeologist, or just push Kotlin maintainers to fix kotlin-scripting implementation in 1.5.32
6) Better Unit tests coverage would be a nice addition to the new implementation.

All the above points should be quite simple to implement. I can pick up point 4 in the nearest future. 
For the other points, we can just create tickets - they are simple enough, so maybe someone can help with implementation.

If you can add me as a committer that would be definitely helpful.

holgerbrandl:
Great. I've added your PR as repo branch https://github.com/holgerbrandl/kscript/tree/kscript4 and invited you to become a maintainer. 

Just let me know where I could/should help.

@ligee Is there a youtrack ticket regarding (5) we could upvote?

ligee:
@holgerbrandl, @aartiPl I've just found the comment here - https://youtrack.jetbrains.com/issue/KT-49511#focus=Comments-27-5467398.0-0, and created an issue - https://youtrack.jetbrains.com/issue/KT-50069. But please create issues yourself next time, comments to the mostly irrelevant issues could get lost very easily.

About the issue itself - I wrote a comment there, but basically you probably need to switch to the new resolving infrastructure.

aartiPl:
@holgerbrandl  - thanks for setting me up. As I mentioned previously I plan to work on point 4 soon. If you have some spare time, you can pick any other point :-)

@ligee - thanks for creating the ticket - I was not sure how to raise it correctly. I have added to the ticket some comments. If you need anything else please just ping me here.

aartiPl:
I have readded dependencies caching - in my simple checks with some particular test case, it was about 1s (~2.6s vs. ~1.6) in favor of the cached version.

holgerbrandl:
With dependncies caching, do you refer to the classpath cachelist file in  ~/.kscript/dependency_cache.txt ?
Indeed there is no test-coverage for it (and it would be tricky to do so as it depends on system and network performance), but it certainly impacts performance. We should preserve this feature, as it is providing a significant performance boost without much effort.

aartiPl:
Yes, I agree. It is restored now on kscript4 branch.

aartiPl:
@holgerbrandl , @ligee - I have updated Kotlin to version 1.6.10 and coroutines to 1.6.0-RC3 and now most of the previously failing tests in the context of resolving dependencies (test/test_suite.sh) are passing. The only failing test is:

`
kscript /home/vagrant/workspace/Kod/Repos/kscript/test/resources/depends_on_dynamic.kts
`

obraz

Can you please check if this test case works for you?

holgerbrandl:
Awesome, thanks @aartiPl . I won't manage right today, but with the Christmas vacation coming soon, I will have more time than usual to focus on this exciting refactoring in the coming days.

holgerbrandl:
Thank you so much for your patience and hard work @aartiPl . I've finally managed to merge the changes and release it as v4.0.0 on all release channels. Usually, migration problems area reported soon after.

Just let me know if you see more appropriate ways to mark/document your great contribution to the project.

holgerbrandl:
Regarding the failing test /resources/depends_on_dynamic.kts
It still fails in the latest version because the resolver seems to struggle with artifact locators such as "log4j:log4j:[1.2,)". These were previously supported. However, I'd consider this a rare edge case, and have published v4 irrespective of that regression.

aartiPl:
I am really glad to hear that! 👍 I will try to support you as much as I can with the release. If you can put a link to my Github page in the acknowledgment section in Readme it would be great :-)

I have noticed there is already a report about resolving artifacts: I suspect it is a problem that I have reported above (I had exactly the same symptoms):
----
I couldn't update kotlin-scripting-jvm to v1.5.31 because in that version there is a bug, which prevents the correct resolution of dependencies if they do not exist in the local repository. So now most of the app is compiled with 1.5.31 (needed e.g. for value classes or list builders), but the resolution of dependencies is done through old kotlin-scripting-jvm v1.4.32. So this is not a perfect solution, but seems to work well enough.
----
UPDATE: Ok. It's not really that good. After manually removing .m2 directory I have again a lot of failing tests. Please recheck on your configurations.
----

It seems to be a problem with Kotlin scripting artifact resolver - it picks up local artifacts but doesn't work with remote repositories.

I will add my comment in the reported ticket.

holgerbrandl:
So far I could not find any indication the dependency versions are the root cause of the reported resolver issues. I've added another integration test in `kscript.app.resolver.IntegrationTests` which run directly in the ide and completes without any issue with the current 1.6.20 versions.

I intentionally did so, because within the IDE it runs with `java` directly, whereas in the shell in runs with `kotlin`. In the past this resulted in different behavior (because of classloader issues in `kotlin`), so I thought it might help to check this here as well.

aartiPl:
Here is a temporary fix for the artifact resolution problem (downgraded kotlin-scripting to 1.4.32):
https://github.com/holgerbrandl/kscript/tree/kscript4_temporary_fix

Without that fix below two tests have failed:
obraz

but with changes, they are working as you can see above. (I have checked that the reason for failure was exactly the problem with artifact resolution).

Please recheck on your side.

holgerbrandl:
I've checked and even got the resolver error once, but I can't reproduce it consistently. The mentioned test runs totally fine

```
foobar
brandl@ubuntu:kscript$ rm -rf .m2/repository/com/beust; kscript --clear-cache
Cleaning up cache...
brandl@ubuntu:kscript$ ${PROJECT_DIR}/test/resources/custom_dsl/mydsl_test_with_deps.kts
[kscript] Resolving com.github.holgerbrandl:kutils:0.12...
[kscript] Resolving com.beust:klaxon:5.5...
foobar
brandl@ubuntu:kscript$ 

```

I've also added another test in https://github.com/holgerbrandl/kscript/blob/master/IntegrationTest345.main.kts which runs directly via `kotlin IntegrationTest345.main.kts` and this seems fine as well.

So I still can't reproduce the problem correctly, that's why I'd be also reluctant to integrate this PR.

aartiPl:
Sure. The issue is not easy to reproduce - I have tried it also myself. It might be e.g. some race condition in kotlin-scripting or even in a library used by it under the hood. The only thing which worked for me is the above change in library versions.

aartiPl:
In fact "IntegrationTest345.main.kts" fails fo me every time, even before resolving "com.beust:klaxon:5.5". Below the screenshot from session with enabled development mode "-d".

obraz

As you can see it throws from "org.eclipse.aether.resolution.DependencyResolutionException" (DependencyResolver.kt:51).

PR #333, opened 2021-11-26
Add native bundled runtime option

ScottPeterJohnson:
So this PR is more of a proof of concept than anything immediately production-ready, but the gist is that I wanted to try using the [Badass Runtime plugin](https://badass-runtime-plugin.beryx.org/releases/latest/) to generate truly self-contained scriptlet executables. Basically, this bundles a stripped-down JRE runtime into a self-extracting shell script. Inefficient ("Hello World" is 20 MB) but potentially very useful for using KScript in situations without known JRE deployment. I'm not really sure of a good name to distinguish this from the existing fat jar packaging yet, but went with the "--native" flag for the moment.

Also since the Capsule project has been dead for a long while now (and seems to have Java 9+ issues), I tried replacing it with the shadow plugin. I admit I don't really understand the differences between the two (since the Capsule docsite is dead) but they appear to do the same thing, more or less?

Let me know if you're interested or foresee difficulties, and I could put some more effort into this PR. Currently it basically works (on Linux) with the simple tests I've run so far.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,12 +39,12 @@ jobs:
       - name: Install kotlin
         run: |
           source "/home/runner/.sdkman/bin/sdkman-init.sh"
-          sdk install kotlin 1.4.31
+          sdk install kotlin 1.6.0
 
       - name: Install gradle
         run: |
           source "/home/runner/.sdkman/bin/sdkman-init.sh"
-          sdk install gradle 6.7
+          sdk install gradle 7.2
 
 
       # https://stackoverflow.com/questions/50104666/gradle-difference-between-test-and-check
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,8 +1,8 @@
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 
 plugins {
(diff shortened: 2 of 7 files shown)


holgerbrandl:
In particular since the already present --package option is causing problems (because of legacy dependencies) I like the idea. But also the inclusion of the JRE allows for an even higher level of self-containedness.

However, since we're in the progress of merging #332  (fyi @aartiPl ) we may want to stall this PR until the merge of 332 is complete. Clearly, this requires some adaptions here, but since 332 is much bigger, it seems the correct approach to me. Would this make sense to you as well?

What's imho missing in this PR is an addition to the test-suite to cover this new mode of operation.

ScottPeterJohnson:
That makes sense! This should be a smaller PR, so it shouldn't be too much work to adapt it to a major refactoring.

ScottPeterJohnson:
Closing; going to simply rewrite these changes against 4.0

aartiPl:
Hi @ScottPeterJohnson! It's great that you want to work on a native packaging option.

Please take into consideration that in the 4.1 branch there is already recreated "--package", so a good name for another option would be needed.

Please recreate PR against the 4.1 branch, since we do not want to make changes in the stable 4.0 branch.

PR #343, opened 2022-04-25
build.gradle.kts: Executable JAR

jayvdb:
This allows easily running using `java -jar ktscript.jar` , especially useful on Windows without needing to rely on the bash script `ktscript`.


holgerbrandl:
Interesting options. Thanks for the contribution. kscript also works in a bash terminal on windows (e.g. git-bash)

PR #348, opened 2022-04-28
Temporary fix for artifact resolution problem

aartiPl:
Temporary fix for artifact resolution problem (downgraded kotlin-scripting to 1.4.32)


aartiPl:
Fixed by properly creating shadow jar.

Issue #349, opened 2022-04-29
4.X Regression: Invalid script does not report error on second run


holgerbrandl:
I've discovered an imho major regression.

```
brandl@ubuntu:kscript$ kscript invalid_script.kts 
[kscript] Resolving log4j:log4j:1.2.14...
[kscript] [ERROR] compilation of scriplet failed
[kscript] [ERROR] Command     : bash -c kotlinc  -classpath /home/brandl/.m2/repository/log4j/log4j/1.2.14/log4j-1.2.14.jar -d '/home/brandl/.kscript/cache/jar_dc3cc43e50a91471262373fddebc753b/scriplet.jar' /home/brandl/.kscript/cache/jar_dc3cc43e50a91471262373fddebc753b/Invalid_script.kts /home/brandl/.kscript/cache/jar_dc3cc43e50a91471262373fddebc753b/Main_Invalid_script.kt
[kscript] [ERROR] Exit Code   : 1   
[kscript] [ERROR] Stdout      : 
[kscript] [ERROR] Stderr      : 
[kscript] [ERROR] /home/brandl/.kscript/cache/jar_dc3cc43e50a91471262373fddebc753b/Invalid_script.kts:14:43: error: expecting ')'
[kscript] [ERROR] println("kscript with annotations rocks!")
[kscript] [ERROR]                                           ^
[kscript] [ERROR] /home/brandl/.kscript/cache/jar_dc3cc43e50a91471262373fddebc753b/Invalid_script.kts:14:1: error: too many arguments for public open fun getRootLogger(): Logger! defined in org.apache.log4j.Logger
[kscript] [ERROR] println("kscript with annotations rocks!")
[kscript] [ERROR] ^
[kscript] [ERROR] .stderr
brandl@ubuntu:kscript$ kscript invalid_script.kts 
[kscript] [ERROR] /home/brandl/.kscript/cache/jar_dc3cc43e50a91471262373fddebc753b/jarArtifact.descriptor

```

Now the same with 3.1
```
brandl@ubuntu:kscript$ sdk default kscript 3.1.0
Default kscript version set to 3.1.0
brandl@ubuntu:kscript$ rm -rf .m2/repository/log4j/log4j/1.2.14/; kscript --clear-cache
Cleaning up cache...
brandl@ubuntu:kscript$ kscript invalid_script.kts 
[kscript] Resolving dependencies...
[kscript] Resolving log4j:log4j:1.2.14...
[...]

PR #350, opened 2022-04-29
bugfix for #349

aartiPl:


holgerbrandl:
Thx. You're a repo maintainer as well, so no need to PR on each bugfix. :-)

PR #354, opened 2022-05-18
Kscript 4.1

aartiPl:
Draft PR - there is still more to do.

@holgerbrandl - Can you please release a patched version of kscript 4.0? That way users can still use version 4, while work on 4.1 can be continued.


holgerbrandl:
Sure, done. Thanks for your great initiative here!

aartiPl:
@wuseal - Unfortunately, current changes won't help. The problem is that compilation happens on a single script consolidated from many smaller scripts. Therefore it is not possible to report correct line numbers for specific scripts. Maybe a better approach would be to provide separate scripts for compilation, the compiler will be able to provide exact line numbers of problems in the code. It is doable but requires considerable rework of the KScript. I will consider it for later :-)

PR #357, opened 2022-05-20
Fix for OsType comparison logic

justixLoL:
Changed equals to contains so run for os 'darwin21' also works


aartiPl (approved):
LGTM - should not break anything and also handle the case for MacOs.

PR #364, opened 2022-06-10
Fix #335: Remove the need for the  Gradle capsule plugin.

### LINKED ISSUE #335: Gradle 7 support
It doesn't work with gradle 7.x due to this [gradle-capsule-plugin issue](https://github.com/danthegoodman/gradle-capsule-plugin/issues/43) 😢

vsajip:
Packaging should now work with Gradle 7.

### DIFF
--- a/src/main/kotlin/kscript/app/code/GradleTemplates.kt
+++ b/src/main/kotlin/kscript/app/code/GradleTemplates.kt
@@ -65,7 +65,6 @@ object GradleTemplates {
 
     fun createGradlePackageScript(script: Script, jarArtifact: JarArtifact): String {
         val kotlinOptions = kotlinOptions(script.compilerOpts)
-
         val kotlinVersion = KotlinVersion.CURRENT
         val extendedDependencies = setOf(
             Dependency("org.jetbrains.kotlin:kotlin-stdlib"),
--- a/src/main/kotlin/kscript/app/code/GradleTemplates.kt
+++ b/src/main/kotlin/kscript/app/code/GradleTemplates.kt
@@ -105,41 +105,23 @@ object GradleTemplates {
         tasks.register("makeScript") {
             dependsOn(":jar")
             doLast {
-                val hdr = layout.projectDirectory.toString()
-                val arc = layout.buildDirectory.file("libs/$baseName.jar").get().toString()
-                val out = layout.buildDirectory.file("libs/$baseName").get().toString()
-                val eol = System.getProperty("line.separator").encodeToByteArray()
-                val p = Paths.get(hdr).resolve("exec_header.sh")
-                val hb = Files.readAllBytes(p)
-                val ab = Files.readAllBytes(Paths.get(arc))
-                val outfile = Paths.get(out).toFile()
-                val fos = FileOutputStream(outfile)
-                fos.write(hb)
-                fos.write(eol)
-                fos.write(ab)
-                fos.close()
+                val headerDir = layout.projectDirectory.toString()
+                val jarFileName = layout.buildDirectory.file("libs/$baseName.jar").get().toString()
+                val outFileName = layout.buildDirectory.file("libs/$baseName").get().toString()
+                val lineSeparator = System.getProperty("line.separator").encodeToByteArray()
+                val headerPath = Paths.get(headerDir).resolve("exec_header.sh")
+                val headerBytes = Files.readAllBytes(headerPath)
+                val jarBytes = Files.readAllBytes(Paths.get(jarFileName))
+                val outFile = Paths.get(outFileName).toFile()
+                val fileStream = FileOutputStream(outFile)
+
+                fileStream.write(headerBytes)
+                fileStream.write(lineSeparator)
+                fileStream.write(jarBytes)
+                fileStream.close()
             }
         }
 
-        /*
-         * Old code here for now, but will be removed soon!
-           Capsule: https://github.com/ngyewch/gradle-capsule-plugin
-
-        tasks.create<us.kirchmeier.capsule.task.FatCapsule>("simpleCapsule") {
-            applicationClass("$capsuleApp")
-            archiveFileName.set("$baseName")
-
-            // https://github.com/danthegoodman/gradle-capsule-plugin/blob/master/DOCUMENTATION.md#really-executable-capsules
-            reallyExecutable
-
-            capsuleManifest.apply {
-                applicationClass = "$capsuleApp"
-                application = "$baseName"
-                applicationScript = "exec_header.sh"
-                jvmArgs = listOf()
-            }
-        }
-         */
         dependencies {
             implementation(files("${jarArtifact.path.parent.resolve("scriplet.jar")}"))
             ${hangingIndents(createGradleDependenciesSection(extendedDependencies), 12)}
--- a/src/main/kotlin/kscript/app/code/GradleTemplates.kt
+++ b/src/main/kotlin/kscript/app/code/GradleTemplates.kt
@@ -39,7 +39,30 @@ object GradleTemplates {
         """.trimIndent()
     }
 
-    //Capsule: https://github.com/ngyewch/gradle-capsule-plugin
+    fun hangingIndents(s: String, n: Int): String {
+        val parts = s.split('\n')
+        val indent = String.format("%${n}s", " ")
+        val result = StringBuilder()
+        val psize = parts.size
+
+        if (psize == 1) {
+            return s
+        }
+        for ((i, p) in parts.withIndex()) {
+            if (i == 0) {
+                result.append(p)
+            }
+            else {
+                result.append(indent)
+                result.append(p)
+            }
+            if (i < (psize - 1)) {
+                result.append('\n')
+            }
+        }
+        return result.toString()
+    }
+
     fun createGradlePackageScript(script: Script, jarArtifact: JarArtifact): String {
         val kotlinOptions = kotlinOptions(script.compilerOpts)
 
@@ -66,7 +89,7 @@ object GradleTemplates {
         repositories {
             mavenLocal()
             mavenCentral()
-            ${createGradleRepositoriesSection(script.repositories).prependIndent()}
+            ${createGradleRepositoriesSection(script.repositories).prependIndent("")}
         }
 
         tasks.jar {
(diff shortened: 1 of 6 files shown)


vsajip:
I tested with Gradle 7.4.2.

vsajip:
OK, there were three packaging tests. I've commented one out (it was looking to execute a non-existent jar file, and we don't do that now, anyway - we execute scripts that contain jars inside them). The other two tests pass after some minor changes.

aartiPl:
Hi! Your code works wonderfully 👍 I think that we should apply it to the branch kscript_4.1 as there is a new development. In this branch, I have added OS independent paths, so testing multi-OS behavior should be easier. There are only a few changes that may affect your patch:

1. All occurrences of Path were changed to OsPath. Generally, all common functionality should be covered in OsPathExt.kt
2. In parsed script fileName is just a file name without extension - you don't have to use dropExtension in your code
3. test_suite.sh -> linux_suite.sh
4. Unit test for CommandResolver still fails - I was working on Windows cmd support, but it is really deeply screwed up, so it takes a lot of time

So please rebase this patch to kscript_4.1. I will add also some comments to your code.

src/main/kotlin/kscript/app/creator/PackageCreator.kt
```diff
@@ -6,28 +6,51 @@ import kscript.app.model.Script
...
+        val parts = mutableListOf("")
+
+        for (opt in script.kotlinOpts) {
+            val s = opt.value
+            if (s.startsWith("-J")) {
+                parts.add(s.substring(2))
+            }
+        }
+        val opts = parts.joinToString(" ").trim()
+        hdr = hdr.replace("java -jar", "java $opts -jar")
```
aartiPl, line 36: It will be better to create a function for the generation of exectuteHeader and pass $opts as an argument to this function than to replace the value textually.

src/main/kotlin/kscript/app/code/GradleTemplates.kt
```diff
@@ -39,53 +39,114 @@ object GradleTemplates {
...
+            configurations["compileClasspath"].forEach { file: File ->
+                from(zipTree(file.absoluteFile))
+            }
+            duplicatesStrategy = DuplicatesStrategy.INCLUDE
+        }
+
+        tasks.register("makeScript") {
+            dependsOn(":jar")
+            doLast {
+                val hdr = layout.projectDirectory.toString()
```
aartiPl, line 108: Please use more descriptive names for variables. Is it possible to simplify this code?

src/main/kotlin/kscript/app/creator/PackageCreator.kt
```diff
@@ -6,28 +6,51 @@ import kscript.app.model.Script
...
+        FileUtils.createFile(basePath.resolve("exec_header.sh"), hdr)
         FileUtils.createFile(
             basePath.resolve("build.gradle.kts"), GradleTemplates.createGradlePackageScript(script, jarArtifact)
         )
 
         executor.createPackage(basePath)
 
-        basePath.resolve("build/libs/appName").toFile().setExecutable(true)
+        val ef = basePath.resolve("build/libs/$baseName").toFile()
```
aartiPl, line 44: Please use descriptive names -> executableFile

src/main/kotlin/kscript/app/creator/PackageCreator.kt
```diff
@@ -6,28 +6,51 @@ import kscript.app.model.Script
...
+        ef.setExecutable(true)
+
+        infoMsg("Finished packaging '${script.scriptName}'; executable path: ${basePath}/build/libs/$baseName")
```
aartiPl, line 47: This message is not printed if the package is already in cache. Can we print the executable path every time, even if it comes from the cache? As we do not copy file (which I think is okay) we should at least communicate full executable path to the user.

Currently, this is printed whenever new package is created:
[kscript] Packaging script 'url_test.kts' into standalone executable...
[kscript] Finished packaging 'url_test.kts'; executable path: /home/vagrant/.kscript/cache/package_5a501be28df8e8a4ace2de5f8162d0ee/build/libs/url_test
[kscript] Package created in: /home/vagrant/.kscript/cache/package_5a501be28df8e8a4ace2de5f8162d0ee

And only this if it comes from cache:
[kscript] Package created in: /home/vagrant/.kscript/cache/package_5a501be28df8e8a4ace2de5f8162d0ee

aartiPl (changes requested):

src/main/kotlin/kscript/app/code/GradleTemplates.kt
```diff
@@ -39,53 +39,114 @@ object GradleTemplates {
         """.trimIndent()
     }
 
-    //Capsule: https://github.com/ngyewch/gradle-capsule-plugin
+    fun hangingIndents(s: String, n: Int): String {
```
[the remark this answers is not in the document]
vsajip: `trimMargin()` trims margins. This function _adds_ a margin of _n_ spaces on all except the first line. This is so that multi-line dependencies look nice in the output. It's purely cosmetic, but with it, the `dependencies` section in the package's `build.gradle.kts` looks something like this:
```
dependencies {
    implementation(files("/home/vinay/.kscript/cache/jar_db71153912c566b4d192ee7f36d85a8a/scriplet.jar"))
    implementation("org.jetbrains.kotlin:kotlin-stdlib")
    implementation("org.jetbrains.kotlin:kotlin-script-runtime:1.6.21")
    implementation("com.github.ajalt.clikt:clikt:3.4.2")
}
```
Without it, you get
```
        dependencies {
            implementation(files("/home/vinay/.kscript/cache/jar_cfdf0ec49939ebcc9ca0b9a9cbe98f4c/scriplet.jar"))
            implementation("org.jetbrains.kotlin:kotlin-stdlib")
implementation("org.jetbrains.kotlin:kotlin-script-runtime:1.6.21")
implementation("com.github.ajalt.clikt:clikt:3.4.2")
        }
```
So functionally the same, but looks uglier.

src/main/kotlin/kscript/app/code/GradleTemplates.kt
```diff
@@ -39,53 +39,114 @@ object GradleTemplates {
...
         val kotlinOptions = kotlinOptions(script.compilerOpts)
-
         val kotlinVersion = KotlinVersion.CURRENT
         val extendedDependencies = setOf(
             Dependency("org.jetbrains.kotlin:kotlin-stdlib"),
             Dependency("org.jetbrains.kotlin:kotlin-script-runtime:$kotlinVersion")
         ) + script.dependencies
 
         val capsuleApp = jarArtifact.execClassName
+        val baseName = script.scriptName.dropExtension()
```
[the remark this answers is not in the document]
vsajip: Thanks, will bear in mind when I merge to that

src/main/kotlin/kscript/app/creator/PackageCreator.kt
vsajip: Not sure _exactly_ what you mean, but I'll have a go.

src/main/kotlin/kscript/app/creator/PackageCreator.kt
vsajip: The current behavior is unintuitive when the package is already in cache. The exact intended behavior of `--package` is unclear - with a clean cache, the packaged executable is placed in the current directory at the end. With a populated cache, it isn't (even if it is absent) - this doesn't seem intuitive. I would expect the following behavior:
```
if (package not in cache) {
  create package in cache
  build package using Gradle
  report success or failure of build
}
if (built executable in cache) {
  copy executable to current directory
}
report package location and executable path in cache
```
Isn't this better than what we have now?

src/main/kotlin/kscript/app/code/GradleTemplates.kt
```diff
@@ -39,53 +39,114 @@ object GradleTemplates {
...
+                val fos = FileOutputStream(outfile)
+                fos.write(hb)
+                fos.write(eol)
+                fos.write(ab)
+                fos.close()
+            }
+        }
+
+        /*
+         * Old code here for now, but will be removed soon!
```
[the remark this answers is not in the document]
vsajip: Done.

src/main/kotlin/kscript/app/creator/PackageCreator.kt
vsajip: Done.

src/main/kotlin/kscript/app/code/GradleTemplates.kt
vsajip: Done.

src/main/kotlin/kscript/app/creator/PackageCreator.kt
vsajip: Done.

vsajip:
OK, I've addressed most of the comments and added questions where things are unclear. Please take a look, and I will then look at merging with 4.1 in a separate PR.

src/main/kotlin/kscript/app/creator/PackageCreator.kt
aartiPl: I am not a big fan of copying files around - I would rather clearly communicate to the user where is the resulting file. Then it will be just enough to check in Cache.getOrCreatePackage() if the package file is already there or not. Currently, there is only a check for directory - if it exists then it is assumed that the package exists, which might be the wrong assumption. So definitely cache condition has to be improved: if there is a directory, but no package, PackageCreator should be started again. From my point of view, I would rather not copy files to the current directory, but in all cases (cached/not cached) communicate to user where is package located.

aartiPl: An additional advantage of not copying files in 4.1 will be that PackageCreator would be testable with only Junit Tests. With OsPath there are mostly no references to the underlying system, except for some utility classes, which possibly can be mocked.

aartiPl (changes requested):
Please rebase on 4.1 branch - I think we shouldn't introduce any more breaking changes in 4.0

vsajip:
> Please rebase on 4.1 branch - I think we shouldn't introduce any more breaking changes in 4.0

Done, see #370. Note that `--package` was already broken, hence this issue.

vsajip:
> From my point of view, I would rather not copy files to the current directory, but in all cases (cached/not cached) communicate to user where is package located.

Sure, but that's a change to existing behavior, isn't it? For example, the packaging tests expect the executable to be placed in the current directory. Of course, they could be changed, but the cache package location for the executable could change every time because of hash changes. So I'm not the current simplistic Bash-based tests will cope ...

vsajip:
> I think it is not worth adding an additional indentation function

OK, I'll see about removing it.

PR #366, opened 2022-06-17
Fix #302: Handle paths to local jars in dependency declarations.

### LINKED ISSUE #302: Support local jar references in @file:DependsOn
I'm struggling to understand how I'm supposed to get access to my local libraries in kscript. I have  simple file :

    #! /usr/bin/env kscript
    import se.explodera.util.sql.OracleDBFileLoader
    println("Hello.")

where the imported class can be found in a jar file on the environment classpath. But when I run the script I get the message

    error: unresolved reference: se
    import se.explodera.util.sql.OracleDBFileLoader

Do I really have to add my local jar file to a maven repository?

When I try to print the classpath from within my code it seems to be limited to `kotlin-runner.jar`. Is this by design, or am I doing something wrong?

vsajip:


vsajip:
I've verified that this works with my packaging PR #364 , i.e. local jar dependencies are correctly added when creating a package in Gradle.  I can add tests for this as a separate, later PR (as it needs both this PR and #364 to be merged to be able to test both packaging and non-packaging scenarios).

aartiPl:
Closing it as per discussion in https://github.com/holgerbrandl/kscript/issues/302.
Another solution is mentioned in this ticket, which IMHO would be a much better approach.

Issue #367, opened 2022-06-20
Why the need to pass $OSTYPE on the command line?


vsajip:
Is it not possible to determine the OS the system is running on without this?

aartiPl:
As far as I know, It is not possible to find out if the system is Cygwin or MSys, but for those systems, paths have to be converted to correctly start the bash process. Otherwise, we can detect the system type.

vsajip:
The following script:
```kotlin
import java.io.BufferedReader
import java.io.IOException
import java.util.concurrent.TimeUnit

fun whatsMyOS(): String? {
    val cmd = "uname -o".split(" ")
    try {
        val proc = ProcessBuilder(*cmd.toTypedArray())
                .redirectOutput(ProcessBuilder.Redirect.PIPE)
                .redirectError(ProcessBuilder.Redirect.PIPE)
                .start()

        proc.waitFor(60, TimeUnit.MINUTES)
        val stdout = proc.inputStream.bufferedReader().readText()
        return stdout.replace("GNU/", "").toLowerCase()
    } catch(e: IOException) {
        e.printStackTrace()
        return null
    }
}

println(whatsMyOS())
```
Should print `linux`, `darwin`, `cygwin`, `msys` on the various POSIX systems. Also, why pass it as a command line argument when it could be picked up as an environment variable?

aartiPl:
Unfortunately, it won't work for Windows as there is no 'uname' command. Also calling external processes is quite heavy and will consume additional time for kscript execution (which is not desired for scripting solution).

In regards to passing value by env variable: this OSTYPE env is somewhat special in that regard: it can not be read by normal Java way of reading env variables. Seems that it is specific to the bash environment and is not like the real env variable, so basically this method doesn't work.

What's wrong with passing OSTYPE as kscript argument? Do you see any issues with that?

vsajip:
> Unfortunately, it won't work for Windows as there is no 'uname' command

But `uname` need only be invoked to distinguish between `cygwin` and `msys`, right? 

> What's wrong with passing OSTYPE as kscript argument? Do you see any issues with that?

Well, it clearly works, but it's just ugly, and AFAIK no other cross-platform scripting environment seems to need such an approach.

[...]

PR #369, opened 2022-06-20
Pass OSTYPE in environment rather than command-line argument.

vsajip:
This approach appears to work fine in Bash on Ubuntu, macOS, Cygwin and MSYS2. I haven't tested on Windows (other than Cygwin/MSYS2), but I wouldn't expect any problems.


aartiPl:
It was discussed here: https://github.com/holgerbrandl/kscript/issues/367
I am closing it as it is not the best fit for the current design.

PR #370, opened 2022-06-21
Fix #335: Get --package working with Gradle 7 by replacing the capsule plugin

### LINKED ISSUE #335: Gradle 7 support

vsajip:
This has now been merged with the `kscript_4.1` branch.


aartiPl:
Great job - thanks again!

Are you able to still work on tests, cache, and a message with the location of the created package? Please let me know.

If you can:
* Cache should check not only for a directory but also for the existence of the package file itself. That will cover cases where build of the package has failed.
* The message with location of the package should be printed nevertheless if it comes from build or from cache
* Tests should be adjusted, so that they are using printed package location to do the checks.

aartiPl:
Okay - I understand: live 😀

I will leave this PR open for about 2 weeks. If you have some spare time, you are welcome to continue working on PR. If anyone else wants to work on it - please add a comment here. Otherwise, I will merge the PR into the 4.1 branch and fix the remaining issues later.

Issue #323, opened 2021-08-10
Respect XDG Spec


xeruf:
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
Please do not litter my home directory with a `.kscript` folder.

holgerbrandl:
You could simply set a KSCRIPT_CACHE_DIR environment variable to whatever directory you prefer.

What would be a better default?

xeruf:
It is only cache? Well that's easy, `KSCRIPT_CACHE_DIR` should default to `${XDG_CACHE_HOME:-$HOME/.cache}/kscript`

holgerbrandl:
I've checked on ubuntu 20.04 but there is no XDG_CACHE_HOME. In case it's not a standard Linux kernel/shell features, I guess the user should configure it herself, by setting KSCRIPT_CACHE_DIR as described.

xeruf:
Please take a look at the spec. If the variable is not set, it should default to `$HOME/.cache` as I said above. It has been an accepted convention for over a decade, because people do not want random programs littering their HOME.

aartiPl:
@xeruf - I have seen your patch (thanks a lot for that!) and have a few comments about it:

1) Please rebase your patch against kscript_4.1; Its development branch and master should just contain bugfixes.
2) In the long term, I think there might be a configuration file for kscript. Since that, there was AppDir abstraction, where inside resides cache, but there can also be configuration file; What should be config file location according to freedesktop.org specification? Can you add even a simple config file handling in your patch?
3) kscript is intended to run on all OSes: linux, mac, msys, cygwin and other posix, but also on windows. Where should we locate cache/configuration file on Windows? Some tests for cache/config will be necessary for all OSes (I am working right now on increasing the portability of kscript, which is actually not that easy).

If you need some help on how to run tests - please let me know.

xeruf:
XDG spec can be used on any platform.
I am not aware of having submitted any patch here?

aartiPl:
@xeruf  - you are right - the patch comes from @meztihn. Anyway, I think it should be discussed here.

meztihn:
@aartiPl
2. The directory would be `$XDG_CONFIG_HOME` or `$HOME/.config` with our config file. `kscript.conf` for example. What type of config file we want? I am afraid that anything other than `.properties` would require a dependency for deserialization.
[...]

PR #372, opened 2022-07-09
Replace custom .kscript dir with XDG_CACHE_HOME (#323)

meztihn:


aartiPl:
@meztihn - thanks for the patch. Please see https://github.com/holgerbrandl/kscript/issues/323 for comments.

PR #374, opened 2022-07-16
Replace .kscript dir with OS specific dirs (#323)

meztihn:
Recreated PR since I can't change the source branch.
An unrelated note, since sdkman is recommended as a package manager: it works only with bash, but IDEA starts from sh by default. This leads to a problem with missing env variables for packages installed through sdkman (e.g. gradle, kotlin). Developers of sdkman [won't do antyhing about it](https://github.com/sdkman/sdkman-cli/issues/568), so the easiest workaround is to edit idea.sh and change it's shebang from sh to bash. It's pretty sneaky problem and it's description could be useful somewhere in readme for contributors.


aartiPl:
Hi @meztihn! 
The patch looks really great - I basically can not find anything which could be improved.
Thanks a lot for your effort! 

I will test the patch locally on my computer and merge it to the 4.1 branch very soon.

PR #390, opened 2023-01-21
Fix calling kscript.bat by symlink and PowerShell compatibility

goto1134:
The scenario:
1. Unpack kscript files to a folder which is not in the PATH
2. Make a symlink to kscript.bat
3. Run `kscript "println('1')"` in cmd or `kscript """println('1')"""` in PowerShell.

You will get the following error:
```
Error: Could not find or load main class io.github.kscripting.kscript.KscriptKt
Caused by: java.lang.ClassNotFoundException: io.github.kscripting.kscript.KscriptKt
```

The reason is that kscript.bat strictly depends on kscript.bat being in your PATH. It uses `where` (`where.exe` from `System32`)  to locate the kscript.bat in the PATH. 

Another problem with this approach is that when executed in PowerShell,  `where` becomes an alias to the `Where-Object` PowerShell function, which returns an empty string.
 
The provided solution both makes symlinks work and fixes compatibility with PowerShell.

### DIFF
--- a/src/kscript.bat
+++ b/src/kscript.bat
@@ -55,7 +55,7 @@ rem ##########################################################################
 rem # subroutines
 
 :set_home
-  set _BIN_DIR=%~dp0
+  set _BIN_DIR=%~dps0
   set _KOTLIN_HOME=%_BIN_DIR%..
 goto :eof
 
--- a/src/kscript.bat
+++ b/src/kscript.bat
@@ -45,8 +45,7 @@ if !_java_major_version! geq 9 (
   set JAVA_OPTS=!JAVA_OPTS! "--add-opens" "java.base/java.util=ALL-UNNAMED"
 )
 
-for /f "tokens=* USEBACKQ" %%o in (`where kscript.bat`) do set ABS_KSCRIPT_PATH=%%o
-set CLASS_PATH=%ABS_KSCRIPT_PATH:~0,-16%\bin\*
+set CLASS_PATH=%_BIN_DIR%\*
 
 java !JAVA_OPTS! -classpath %CLASS_PATH% io.github.kscripting.kscript.KscriptKt windows %KOTLIN_OPTS%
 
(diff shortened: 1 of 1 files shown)


aartiPl:
Hi! First of all, thanks for providing the PR. I am happy to merge it. My only concerns are connected with the options 's' and 'i', which were available in
```
for %%i in (%~sf0) do set _BIN_DIR=%_BIN_DIR%%%~dpsi
```
's' is for short file names - I guess it should stay, but I am not sure about 'i'.
Have you been able to check those options?

goto1134:
s is for short names (https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#short-vs-long-names)
i is for the loop variable

As I understand the script

```
for %%i in (%~sf0) do set _BIN_DIR=%_BIN_DIR%%%~dpsi
```
can be translated to kotlin-like pseudocode this way:
```
var _BIN_DIR = ""
for (i in listOf(getShortExecutableFileName)) {
  _BIN_DIR = getShortParentDirectoryPath(i);
}
```
The whole loop is equivalent to `%~dps0`. I will make a change to add `s` to the expression. But I must say that I do not know if this change affects the script result anyhow.

aartiPl:
@goto1134 - I have merged the PR. 

If you would like to work on something regarding Windows compatibility, I will gladly welcome the patches. For example distribution package for windows is needed: scoop or choco (mostly investigation first). Other than that, more tests in all areas are always welcome.

Thanks a lot for your contribution!

aartiPl:
That sounds cool! I have created an issue for that:
https://github.com/kscripting/kscript/issues/391

It would be great if you could document in a simple way how to start and, later, how to do releases. Then I will be able to create release scripts. I have no idea what should be done, so you can probably help also in other ways. :-)

goto1134:
Actually you don't need to do anything specific. Just stick to the current GitHub release format and scoop should automatically update the package version once in a while.

If you are interested in how scoop works, you should check its [wiki](https://scoop-docs.vercel.app/docs/concepts/Creating-an-app-manifest.html).

aartiPl:
Great! I am working on the release scripts based on kscript, but they need to be more stable, to ask you for help with them. For now, I will reference you in the ticket and add a link to the wiki. You are welcome to add additional comments to the ticket.

PR #403, opened 2023-06-22
Fix and expand manual tests with auth repo

rocketraman:
The manual tests for an authenticated repo were not complete.

Fixed the test artifactory server startup.

Added a test for environment variable substitution (test 4). Note that this test follows the current documentation in the README of using `{{FOO}}` style env var substitutions, but that will not currently work. What will work is `$FOO`. This should be updated as part of the fix for https://github.com/kscripting/kscript/issues/402.

Added a test for environment variable substitution with packaging (test 5). This test will not currently work as the substituted credentials are not passed to Gradle properly by kscript.

NOTE: Instead of this being a manual test, it could be done in an automated unit test with Testcontainers.


rocketraman:
Closing, will submit a single PR with all related changes.

